PR #32, opened 2025-09-15
enhance: some cells pinning performance optimization

sparknack:
issue: https://github.com/milvus-io/milvus/issues/41435

### DIFF
--- a/include/cachinglayer/CacheSlot.h
+++ b/include/cachinglayer/CacheSlot.h
@@ -34,7 +34,7 @@
 #include "common/OpContext.h"
 #include "log/Log.h"
 
-    namespace milvus::cachinglayer {
+namespace milvus::cachinglayer {
 
 template <typename CellT>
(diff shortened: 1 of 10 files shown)


include/cachinglayer/CacheSlot.h
```diff
@@ -87,11 +87,18 @@ class CacheSlot final : public std::enable_shared_from_this<CacheSlot<CellT>> {
             cids.push_back(i);
         }
         // TODO: Warmup is not tracked for now
-        SemiInlineGet(PinCells(nullptr, cids));
+        PinCellsDirect(nullptr, cids);
```
zhengbuqian, line 90: comment at `Warmup`: this method should not be called in a multi-thread condition. If we want to call `Warmup`, no other methods of `CacheSlot` can be called until the warm up is finished

zhengbuqian (approved):

PR #40, opened 2025-09-28
Add tracer test

chasingegg:
Move UT from https://github.com/milvus-io/milvus/blob/master/internal/core/unittest/test_tracer.cpp


test/test_tracer.cpp
```diff
@@ -0,0 +1,238 @@
+#include <gtest/gtest.h>
```
zhengbuqian, line 1: how about TracerTest.cpp

test/TracerTest.cpp
```diff
@@ -0,0 +1,238 @@
+#include <gtest/gtest.h>
```
chasingegg: done

PR #69, opened 2026-03-10
Migrate to Conan 2.x

yhmo:
new JFrog artifactory for conan 2.x: https://milvus01.jfrog.io/ui/repos/tree/General/default-conan-local2

issue: https://github.com/milvus-io/milvus/issues/47425

conanfiles pr: https://github.com/milvus-io/conanfiles/pull/92

### DIFF
--- a/.github/workflows/ut.yaml
+++ b/.github/workflows/ut.yaml
@@ -10,7 +10,10 @@ on:
     paths:
       - 'include/**'
       - 'src/**'
-      - 'tests/**'
+      - 'test/**'
+      - 'conanfile.py'
+      - 'Makefile'
+      - 'CMakeLists.txt'
       - '.github/workflows/ut.yaml'
 
 jobs:
@@ -18,6 +21,9 @@ jobs:
     name: UT on ubuntu-22.04
     runs-on: ubuntu-22.04
     timeout-minutes: 240
+    env:
+      CC: gcc-12
+      CXX: g++-12
     strategy:
       fail-fast: false
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/ut.yaml
```diff
@@ -28,21 +28,26 @@ jobs:
       - name: Install Dependency
         run: |
           sudo apt update && sudo apt install -y cmake libopenblas-dev libaio-dev g++-12 gcc-12 python3 python3-pip \
-          && pip3 install conan==1.65.0 \
-          && conan remote add default-conan-local https://milvus01.jfrog.io/artifactory/api/conan/default-conan-local
+          && pip3 install conan==2.25.1 \
```
liliu-z: paths filter uses 'tests/**' but the repo directory is 'test/**', so this entry never matches. Also missing conanfile.py, Makefile, and CMakeLists.txt, which means build/dependency changes won't trigger UT.

.github/workflows/ut.yaml
```diff
@@ -28,21 +28,26 @@ jobs:
...
+          && conan profile detect \
```
liliu-z: Cache restore runs before 'conan profile detect --force' and 'conan remote add --force', allowing stale cache to silently overwrite profile/remotes (e.g. Conan 1 remote default-conan-local on first migration run).

README.md
```diff
@@ -17,10 +17,11 @@ Milvus Common is a core component library that serves as a bridge between Milvus
 ## Build from source
 
 ```
-pip install conan==1.64.0
+pip install conan==2.25.1
+conan profile detect
```
liliu-z, line 21: Missing 'conan remote add default-conan-local2 ...' step; docs are not reproducible on a clean machine.

.github/workflows/ut.yaml
liliu-z: CC/CXX not set: conan profile detect picks gcc 11 on Ubuntu 22.04 while CONAN_ARGS forces -s compiler.version=12, causing package ID to be computed for gcc 12 while actually compiling with gcc 11 (silent ABI mismatch).

Makefile
```diff
@@ -27,7 +27,7 @@ conan: deps
 	$(SDKROOT_ENV) \
```
liliu-z, line 27: Missing Conan 2 bootstrap steps. Add `conan profile detect --force` and `conan remote add ... --force` for CI parity and idempotency.

README.md
```diff
@@ -17,10 +17,12 @@ Milvus Common is a core component library that serves as a bridge between Milvus
 ## Build from source
```
liliu-z, line 17: Conan setup not aligned with CI: missing `--force` on both `profile detect` and `remote add` (non-idempotent, fails on re-run), missing `compiler.version=12`, and no mention of gcc-12 requirement.

conanfile.py
```diff
@@ -1,34 +1,31 @@
...
 from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
 from conan.tools.gnu import PkgConfigDeps
 from conan.errors import ConanInvalidConfiguration
-from conans import tools
 import os
 
-required_conan_version = ">=1.55.0"
-
 class MilvusCommonConan(ConanFile):
     settings = "os", "compiler", "build_type", "arch"
```
liliu-z, line 14: gtest/1.15.0 is not pinned to a revision hash, risking dependency drift.

conanfile.py
```diff
@@ -37,26 +34,30 @@ class MilvusCommonConan(ConanFile):
...
     def requirements(self):
         # Force all dependencies to use protobuf from milvus/dev channel
         # This is needed to resolve conflicts between opentelemetry-cpp and grpc
-        self.requires("protobuf/5.27.0@milvus/dev#6fff8583e2fe32babef04a9097f1d581", force=True, override=True)
-        self.requires("lz4/1.9.4", force=True, override=True)
+        self.requires("protobuf/5.27.0@milvus/dev#42f031a96d21c230a6e05bcac4bdd633", force=True, override=True)
+        self.requires("lz4/1.9.4#7f0b5851453198536c14354ee30ca9ae", force=True, override=True)
+        # Force overrides openssl to resolve opentelemetry-cpp's transitive deps
+        self.requires("openssl/3.3.2#9f9f130d58e7c13e76bb8a559f0a6a8b", force=True, override=True)
+        self.requires("libcurl/8.10.1#a3113369c86086b0e84231844e7ed0a9", force=True, override=True)
```
liliu-z, line 57: `force=True` without `override=True` on nlohmann_json is semantically inconsistent and makes it a direct dependency of milvus-common, potentially altering downstream link graph and ABI/API contract. Clarify intent and add `override=True` or a comment documenting it as a direct dep.

Makefile
yhmo: Adding conan profile detect --force to every make conan run would overwrite any custom profile the user has configured.                                                                                      
  
The README already includes conan profile detect as a one-time setup step before building. I think that's sufficient.

conanfile.py
yhmo: no need to change

README.md
yhmo: no need to change

Makefile
yhmo: no need change

liliu-z (commented):
/lgtm
/approve

PR #78, opened 2026-03-31
fix: integrate LoadingOverheadTracker into DList to eliminate TOCTOU

sparknack:
Move tracker Reserve/Release calls inside DList's list_mtx_ lock,
making tracker state updates atomic with DList resource reservation.
This eliminates the TOCTOU race where a failed DList reservation
could cause tracker/DList accounting divergence under concurrency.

Key changes:
- DList holds shared_ptr<LoadingOverheadTracker>, set via SetLoadingOverheadTracker()
- New DList overloads: ReserveLoadingResourceWithTimeout(loaded, overhead, handle, ...)
  returns SemiFuture<ResourceUsage> (zero = failure, non-zero = actual reserved)
- ReleaseLoadingResource(loaded, overhead, handle) releases atomically
- LoadingOverheadTracker uses string-based group registration returning uint64_t
  handle for O(1) hot-path lookup; translator controls grouping via Meta
- CacheSlot no longer calls tracker directly, only passes overhead_handle_ to DList
- Manager/CacheSlot tracker pointer replaced with shared_ptr (fixes UAF risk)
- Translator::Meta gains loading_overhead_group field for custom grouping
- WaitingRequest carries overhead_handle for deferred fulfillment

### DIFF
--- a/include/cachinglayer/CacheSlot.h
+++ b/include/cachinglayer/CacheSlot.h
@@ -58,7 +58,7 @@ class CacheSlot final : public std::enable_shared_from_this<CacheSlot<CellT>> {
     CacheSlot(std::unique_ptr<Translator<CellT>> translator, internal::DList* dlist, bool evictable, bool self_reserve,
               bool storage_usage_tracking_enabled, std::chrono::milliseconds loading_timeout,
               std::chrono::milliseconds warmup_loading_timeout,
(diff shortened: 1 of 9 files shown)


zhengbuqian (approved):

chasingegg (approved):
/lgtm

PR #79, opened 2026-04-15
[WIP] feat: add shared io_uring context pool support

CLiqing:
Introduce UringContextPool in milvus-common and gate it behind WITH_IO_URING with liburing detection so downstream knowhere/cardinal can share io_uring contexts safely.


src/knowhere/uring_context_pool.cc
```diff
@@ -0,0 +1,138 @@
+#ifdef WITH_IO_URING
```
liliu-z, line 1: push() no-ops when stop_==true but never removes the ring from checked_out_rings_. The destructor spins waiting for that set to drain, then force-deletes all rings — any thread still holding a pointer gets a deterministic use-after-free.

src/knowhere/io_reader.cc
```diff
@@ -0,0 +1,192 @@
...
+                                  std::vector<struct iocb*> cb_ptrs(buffers.size());
+                                  std::vector<struct io_event> events(buffers.size());
+
+                                  for (size_t i = 0; i < buffers.size(); ++i) {
+                                      io_prep_pread(&cbs[i], fd, reinterpret_cast<void*>(buffers[i]), size, offsets[i]);
+                                      cb_ptrs[i] = &cbs[i];
+                                  }
+
+                                  const auto submitted = io_submit(ctx, static_cast<long>(cb_ptrs.size()), cb_ptrs.data());
+                                  if (submitted != static_cast<long>(cb_ptrs.size())) {
```
liliu-z: On partial submit or I/O failure, the context is returned to the pool without draining in-flight operations. The next user of this context will inherit stale/corrupted CQE state. Same issue at lines 145-147, 153-155, and 158-160.

src/knowhere/io_reader.cc
```diff
@@ -0,0 +1,192 @@
...
+                                          sqe->user_data = idx;
+                                      }
+
+                                      if (batch == 0) {
+                                          pool->PushUring(ring);
+                                          return false;
+                                      }
+
+                                      const auto submitted = io_uring_submit(ring);
+                                      if (submitted < 0 || static_cast<size_t>(submitted) != batch) {
```
liliu-z: Dirty context returned to pool on failure without cleanup — same issue as line 94.

src/knowhere/io_reader.cc
```diff
@@ -0,0 +1,192 @@
...
+                                          pool->PushUring(ring);
+                                          return false;
+                                      }
+
+                                      size_t completed = 0;
+                                      while (completed < batch) {
+                                          io_uring_cqe* cqe = nullptr;
+                                          if (io_uring_wait_cqe(ring, &cqe) < 0 || cqe == nullptr) {
```
liliu-z: Dirty context returned to pool on failure without cleanup — same issue as line 94.

src/knowhere/io_reader.cc
```diff
@@ -0,0 +1,192 @@
...
+                                              pool->PushUring(ring);
+                                              return false;
+                                          }
+                                          if (cqe->res < 0 || static_cast<size_t>(cqe->res) != size) {
+                                              io_uring_cqe_seen(ring, cqe);
```
liliu-z: Dirty context returned to pool on failure without cleanup — same issue as line 94.

src/knowhere/aio_context_pool.cc
```diff
@@ -1,13 +1,14 @@
 #include "knowhere/aio_context_pool.h"
```
liliu-z, line 1: GetGlobalAioPool() silently creates a second, independent AIO pool when the unified pool already selected io_uring. Two global pools with different configs coexist, violating the shared-pool invariant.

src/knowhere/uring_context_pool.cc
```diff
@@ -0,0 +1,138 @@
...
+bool
+UringContextPool::InitGlobalUringPool(size_t num_ctx, size_t max_entries) {
+    IOContextPoolConfig cfg;
+    cfg.prefer_io_uring = true;
+    cfg.num_ctx = num_ctx;
+    cfg.max_events = max_entries;
+    return IOContextPool::InitGlobal(cfg);
+}
+
+std::shared_ptr<UringContextPool>
```
liliu-z, line 237: Same split-brain issue: legacy fallback path can instantiate a separate uring pool alongside the unified pool.

src/knowhere/io_context_pool.cc
```diff
@@ -0,0 +1,194 @@
+#include "knowhere/io_context_pool.h"
```
liliu-z, line 1: InitGlobal() returns true even when the pool is already initialized with a different config. Callers believe their num_ctx/max_events took effect when they didn't. Should return false or at minimum log the mismatch.

include/knowhere/io_context_pool.h
```diff
@@ -0,0 +1,104 @@
...
+#endif
+
+enum class IOBackend {
+    UNKNOWN,
+    IO_URING,
+    AIO,
+};
+
+struct IOContextPoolConfig {
+    size_t num_ctx = 1;
```
liliu-z: num_ctx defaults to 1, but legacy AIO defaulted to 512 (65536/128). Auto-initialization via GetGlobal() with default config will silently shrink the pool from 512 concurrent I/O contexts to 1, causing severe throughput regression on hot paths.

src/knowhere/io_reader.cc
```diff
@@ -0,0 +1,192 @@
+#include "knowhere/io_reader.h"
```
liliu-z, line 1: The AIO read path does not batch submissions against max_events_per_ctx. Combined with the dirty-context-return bug, partial submits become a correctness problem, not just a performance one.

src/knowhere/aio_context_pool.cc
```diff
@@ -29,7 +30,7 @@ AioContextPool::InitGlobalAioPool(size_t num_ctx, size_t max_events) {
 }
```
liliu-z, line 45: Data race: global_aio_pool_size is read without a lock before the lock is acquired and read again, but it is a plain size_t, not std::atomic. The unsynchronized first read is undefined behavior. Same pattern in InitGlobalAioPoolWithValidation().

src/knowhere/uring_context_pool.cc
CLiqing: Fixed in 25708b1. UringContextPool::push now erases checked_out_rings_ when stop_ is true before returning, so shutdown does not keep stale checked-out pointers.

src/knowhere/io_reader.cc
CLiqing: Fixed in 25708b1. The AIO path now batches by MaxEventsPerCtx, tracks pending submits, and drains in-flight events before returning the context on failure.

src/knowhere/io_reader.cc
CLiqing: Addressed by the same 25708b1 fix in io_reader.cc. Failure paths now drain pending AIO completions before PushAio, so dirty contexts are not returned to the pool.

src/knowhere/io_reader.cc
CLiqing: Addressed by the same 25708b1 fix in io_reader.cc. Failure paths now drain pending AIO completions before PushAio, so dirty contexts are not returned to the pool.

src/knowhere/io_reader.cc
CLiqing: Addressed by the same 25708b1 fix in io_reader.cc. Failure paths now drain pending AIO completions before PushAio, so dirty contexts are not returned to the pool.

src/knowhere/aio_context_pool.cc
CLiqing: Fixed in 25708b1. GetGlobalAioPool now returns nullptr when unified backend is not AIO, and InitGlobalAioPool fails on backend mismatch, preventing split-brain fallback pools.

src/knowhere/uring_context_pool.cc
CLiqing: Fixed in 25708b1. GetGlobalUringPool now returns nullptr when unified backend is not IO_URING, and InitGlobalUringPool fails on backend mismatch, preventing split-brain fallback pools.

src/knowhere/io_context_pool.cc
CLiqing: Fixed in 25708b1. IOContextPool::InitGlobal now rejects re-init with mismatched num_ctx or max_events and returns false with explicit logs.

include/knowhere/io_context_pool.h
CLiqing: Fixed in 25708b1. IOContextPoolConfig::num_ctx now defaults to default_pool_size when libaio is enabled, preserving legacy AIO default concurrency.

src/knowhere/io_reader.cc
CLiqing: Fixed in 25708b1. The AIO read path now batches submissions by MaxEventsPerCtx and handles partial submit and completion loops correctly.

src/knowhere/aio_context_pool.cc
CLiqing: Fixed in 25708b1 by converting global_aio_pool_size and global_aio_max_events to std::atomic<size_t>, removing unsynchronized reads in the double-check pattern.

src/knowhere/uring_context_pool.cc
```diff
@@ -0,0 +1,144 @@
+#ifdef WITH_IO_URING
```
liliu-z, line 1: When all io_uring ring inits fail, UringContextPool sets stop_=true but the constructor still returns a non-null pointer, so TryInitUring incorrectly reports success. The pool != nullptr check in io_context_pool.cc is insufficient; you need to verify ring_bak_ is non-empty or expose an IsUsable() predicate.

src/knowhere/io_reader.cc
```diff
@@ -0,0 +1,240 @@
...
+            throw std::invalid_argument("buffer pointer should not be null");
+        }
+    }
+
+    auto pool = io_pool_ ? io_pool_ : IOContextPool::GetGlobal();
+    if (pool == nullptr || !pool->IsInitialized()) {
+        throw std::runtime_error("IOContextPool is not initialized");
+    }
+
+    return std::async(std::launch::async,
```
liliu-z: std::async(std::launch::async) spawns a new OS thread per call. On hot paths this causes thread explosion; a bounded thread pool or io_uring's native async should be used instead.

src/knowhere/io_context_pool.cc
```diff
@@ -0,0 +1,189 @@
...
+            LOG_ERROR("Global IOContextPool already initialized with num_ctx=%zu, requested=%zu", g_io_pool->num_ctx_, cfg.num_ctx);
+            return false;
+        }
+        LOG_WARN("Global IOContextPool has already been initialized with backend: %s", g_io_pool->BackendName().c_str());
+        return true;
+    }
+
+    auto io_pool = std::shared_ptr<IOContextPool>(new IOContextPool());
+
+#ifdef WITH_IO_URING
```
liliu-z, line 169: When io_uring is detected at compile time it is unconditionally enabled with no runtime fallback to AIO if initialization fails. Combined with the false-success bug in TryInitUring, the system can silently run with a non-functional IO backend. There is also no CMake option to manually disable io_uring.

src/knowhere/uring_context_pool.cc
CLiqing: Fixed in 09b593c. Added UringContextPool::IsUsable() (checks ring_bak_ non-empty) and updated IOContextPool::TryInitUring to fail when pool is unusable. This closes the false-success path when io_uring_queue_init fails for all rings.

src/knowhere/io_reader.cc
CLiqing: Addressed in 09b593c. IOReader::ReadAsync now uses std::launch::deferred, so it does not spawn a new OS thread per call. Added regression test IOContextPoolTest.ReadAsyncShouldBeDeferredFuture.

src/knowhere/io_context_pool.cc
CLiqing: Updated in 09b593c. Kept env-only io_uring detection (no manual CMake switch), and added startup fallback to AIO when io_uring init is unavailable. Also fixed the false-success path via IsUsable() guard. Added regression test IOContextPoolTest.InitShouldFallbackToAioWhenUringUnavailable.

test/IOContextPoolTest.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+TEST(IOContextPoolTest, BackendIsSelectedAtInit) {
+    IOContextPoolConfig cfg;
+    cfg.num_ctx = 2;
+    cfg.max_events = 128;
+
+    ASSERT_TRUE(IOContextPool::InitGlobal(cfg));
+
+    auto pool = IOContextPool::GetGlobal();
+    ASSERT_NE(pool, nullptr);
```
liliu-z, line 118: Tests share the global singleton without any reset mechanism, making results depend on GTest execution order. A ResetGlobalForTest() helper or fork()-based isolation (like the fallback test) is needed to prevent flaky failures.

include/knowhere/aio_context_pool.h
```diff
@@ -59,6 +61,15 @@ class AioContextPool {
     static std::shared_ptr<AioContextPool>
     GetGlobalAioPool();
 
+    static bool
```
liliu-z, line 169: Wait predicate only checks ctx_q_.size() and never checks stop_. Threads blocked here will deadlock on destruction when the queue is empty because notify_all() cannot break the wait.

src/knowhere/uring_context_pool.cc
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    std::scoped_lock lk(global_uring_pool_mut);
+    if (global_uring_pool_size == 0) {
+        global_uring_pool_size = num_ctx;
+        global_uring_max_entries = max_entries;
+        return true;
+    }
+
+    if (global_uring_pool_size != num_ctx || global_uring_max_entries != max_entries) {
```
liliu-z, line 70: Same problem as AioContextPool: config mismatch is logged but the function still returns true, silently allowing misconfiguration.

include/knowhere/aio_context_pool.h
```diff
@@ -59,6 +61,15 @@ class AioContextPool {
...
+    InitGlobalAioPoolWithValidation(size_t num_ctx, size_t max_events);
+
+    static std::shared_ptr<AioContextPool>
+    GetGlobalAioPoolDirect();
+
+    static void
+    ResetGlobalForTest();
+
     ~AioContextPool() {
```
liliu-z, line 306: stop_ is a non-atomic bool written without holding ctx_mtx_, which is a data race. Also, io_destroy is called before notify_all, so woken threads may access already-destroyed contexts.

include/knowhere/io_reader.h
```diff
@@ -0,0 +1,84 @@
...
+ private:
+    T* data_;
+    size_t size_;
+};
+}  // namespace knowhere_compat
+#endif
+
+#include "knowhere/io_context_pool.h"
+
+#if defined(__cpp_lib_span)
```
liliu-z: IOReaderSpan resolves to different types depending on __cpp_lib_span; using it in public function signatures causes ABI incompatibility across C++17 and C++20 consumers.

include/knowhere/aio_context_pool.h
CLiqing: Fixed in df90ca2.

Change:
- include/knowhere/aio_context_pool.h: introduced Shutdown() to set stop_ while holding ctx_mtx_, then notify_all() after unlocking.
- Destructor now calls Shutdown() before io_destroy(), so waiters observe stop and exit instead of waiting on destroyed contexts.
- push() now no-ops when stop_ is already true.

Verification:
- Added regression test: test/IOContextPoolTest.cpp: LegacyAioPopShouldReturnNullAfterShutdown
- Built and ran: build-test/test/all_tests --gtest_filter='IOContextPoolTestFixture.*'
- Result: 14/14 passed.

include/knowhere/io_context_pool.h
```diff
@@ -0,0 +1,111 @@
...
+    IO_URING,
+    AIO,
+};
+
+struct IOContextPoolConfig {
+#ifdef MILVUS_COMMON_WITH_LIBAIO
+    size_t num_ctx = default_pool_size;
+#else
+    size_t num_ctx = 1;
+#endif
```
liliu-z, line 45: num_ctx falls back to 1 when libaio is absent, so io_uring-only builds get a single ring and a severe concurrency bottleneck. Default should be reasonable regardless of libaio availability.

src/knowhere/uring_context_pool.cc
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    if (g_uring_pool == nullptr) {
+        g_uring_pool = std::shared_ptr<UringContextPool>(new UringContextPool(global_uring_pool_size, global_uring_max_entries));
+    }
+    return g_uring_pool;
+}
+
+bool
+UringContextPool::InitGlobalUringPool(size_t num_ctx, size_t max_entries) {
```
liliu-z, line 217: Same structural bug as the AIO variant: `InitGlobalUringPool(num_ctx, max_entries)` internally delegates to `IOContextPool::GetGlobal()` which uses the default `IOContextPoolConfig`, so the caller-supplied `num_ctx` and `max_entries` are silently ignored. This is a confirmed bug in the new uring API introduced by this PR.

src/knowhere/aio_context_pool.cc
```diff
@@ -24,21 +29,65 @@ AioContextPool::InitGlobalAioPool(size_t num_ctx, size_t max_events) {
...
+    if (g_aio_pool == nullptr) {
+        g_aio_pool = std::shared_ptr<AioContextPool>(new AioContextPool(global_aio_pool_size, global_aio_max_events));
     }
-    static auto pool = std::shared_ptr<AioContextPool>(new AioContextPool(global_aio_pool_size, global_aio_max_events));
-    return pool;
+    return g_aio_pool;
+}
+
+bool
+AioContextPool::InitGlobalAioPool(size_t num_ctx, size_t max_events) {
```
liliu-z, line 62: InitGlobalAioPool() calls GetGlobal() which uses a default IOContextPoolConfig (num_ctx=512) instead of forwarding the caller's global_aio_pool_size parameter. Any non-default num_ctx configuration is silently ignored — 100% regression for customized deployments. Compare with InitGlobalUringPool (uring_context_pool.cc L92-94) which correctly passes parameters. Test L167 will fail in !WITH_IO_URING environments.

Fix: Construct IOContextPoolConfig from the caller's parameters and pass it directly to InitGlobal(cfg).

CLiqing:
Additional large-dataset validation for standard DISKANN search:

## Setup

Validated with `cohere`, `openai`, and `gist` using `index_version=9` DISKANN.

Two local bundles were built for comparison:
- io_uring bundle: current PR path, `IOContextPool` initializes with backend `io_uring`
- AIO bundle: temporary local source switch forcing the legacy AIO path

The temporary AIO switch and local build-only OpenBLAS workaround were restored after benchmarking; no benchmark-only source changes are intended for this PR.

## Search syscall proof

| backend | dataset | io_uring_setup | io_uring_enter | io_setup | io_submit | io_getevents |
|---|---:|---:|---:|---:|---:|---:|
| io_uring | cohere | 512 | 145417 | 0 | 0 | 0 |
| io_uring | openai | 512 | 122389 | 0 | 0 | 0 |
| io_uring | gist | 512 | 141975 | 0 | 0 | 0 |
| AIO | cohere | 0 | 0 | 512 | 15530 | 15530 |
| AIO | openai | 0 | 0 | 512 | 15029 | 15029 |
| AIO | gist | 0 | 0 | 512 | 16361 | 16361 |

This confirms the standard DISKANN search path uses io_uring in the PR path and legacy AIO in the control path.

## Search performance and recall, median of 3 runs

| dataset | backend | QPS median | avg_recall | avg_ncdg |
|---|---|---:|---:|---:|
| cohere | io_uring | 48.5533 | 0.88799 | 0.912738 |
| cohere | AIO | 48.7682 | 0.88717 | 0.911849 |
| openai | io_uring | 38.2560 | 0.91667 | 0.936018 |
| openai | AIO | 38.1430 | 0.91721 | 0.936543 |
| gist | io_uring | 51.4329 | 0.86638 | 0.885165 |
| gist | AIO | 51.4146 | 0.866881 | 0.885877 |

QPS is effectively flat across the three datasets: -0.44%, +0.30%, and +0.04% for io_uring vs AIO.

## Recall determinism check

The initial comparison used separately built indexes for the io_uring and AIO bundles, so the small recall/ncdg differences above were suspected to come from DiskANN build-time nondeterminism rather than the search IO backend.

I then cross-searched existing indexes without rebuilding:

| search backend | index source | dataset | QPS | avg_recall | avg_ncdg |
|---|---|---:|---:|---:|---:|
| io_uring | AIO-built index | cohere | 48.7743 | 0.88717 | 0.911849 |
| AIO | io_uring-built index | cohere | 48.5533 | 0.88799 | 0.912738 |
| io_uring | AIO-built index | openai | 38.143 | 0.91721 | 0.936543 |
| AIO | io_uring-built index | openai | 38.2568 | 0.91667 | 0.936018 |
| io_uring | AIO-built index | gist | 51.4206 | 0.866881 | 0.885877 |
| AIO | io_uring-built index | gist | 51.4314 | 0.86638 | 0.885165 |

Result: recall/ncdg follows the index files exactly, not the search backend. With the same index, switching between io_uring and AIO preserves recall/ncdg for standard DISKANN search.

src/knowhere/aio_context_pool.cc
```diff
@@ -1,13 +1,18 @@
...
+std::shared_ptr<AioContextPool> g_aio_pool;
+}
+
+std::atomic<size_t> AioContextPool::global_aio_pool_size{0};
+std::atomic<size_t> AioContextPool::global_aio_max_events{0};
 std::mutex AioContextPool::global_aio_pool_mut;
 
 bool
-AioContextPool::InitGlobalAioPool(size_t num_ctx, size_t max_events) {
+AioContextPool::InitGlobalAioPoolWithValidation(size_t num_ctx, size_t max_events) {
```
liliu-z, line 15: The validation at line 20 only rejects `max_events > default_max_events` but allows `max_events == 0` through. The function returns `true` (success) and sets `global_aio_max_events = 0`. Subsequent `GetGlobalAioPoolDirect()` constructs a pool where every `io_setup(0, &ctx)` call fails, producing an empty pool. Any `pop()` call will block indefinitely. Note: the unified entry point `IOContextPool::InitGlobal` at `io_context_pool.cc:60` correctly rejects 0, so this only affects direct callers of the newly-exposed helper API declared in `aio_context_pool.h:76`.

src/knowhere/uring_context_pool.cc
```diff
@@ -0,0 +1,164 @@
...
+
+    LOG_WARN("Global UringContextPool has already been initialized with context num: %zu", global_uring_pool_size);
+    return true;
+}
+
+std::shared_ptr<UringContextPool>
+UringContextPool::GetGlobalUringPoolDirect() {
+    std::scoped_lock lk(global_uring_pool_mut);
+    if (global_uring_pool_size == 0) {
+        global_uring_pool_size = 1;
```
liliu-z: The legacy `GetGlobalUringPoolDirect()` constructs `UringContextPool` with `num_ctx=1` by default, while AIO pool defaults to 512 contexts. Under direct legacy usage, this severely limits uring concurrency.

Fix: Align the default num_ctx with AIO pool defaults or use a reasonable default like `default_pool_size`.

src/knowhere/io_reader.cc
```diff
@@ -0,0 +1,465 @@
...
+        cb_ptrs[i] = &cbs[i];
+    }
+
+    size_t submitted_total = 0;
+    size_t retry = 0;
+    while (submitted_total < batch) {
+        const auto submitted =
+            io_submit(ctx, static_cast<long>(batch - submitted_total), cb_ptrs.data() + submitted_total);
+        if (submitted < 0) {
+            if (-submitted == EINTR) {
```
liliu-z, line 152: At io_reader.cc:81-83, io_reader.cc:113-115, and io_reader.cc:261-263, when the syscall returns `EINTR`, `continue` skips `retry++`. Under sustained signal delivery (e.g., from ptrace or profiler), this could loop indefinitely. Practical risk is very low since EINTR from `io_submit`/`io_getevents`/`io_uring_wait_cqe` is rare.

Fix: Either count EINTR toward the retry limit or add a separate EINTR-specific limit.

src/knowhere/io_completion_reader.cc
```diff
@@ -0,0 +1,345 @@
...
+    auto request_id = next_request_id_++;
+    auto [iter, inserted] = pending_requests_.emplace(request_id, RequestState{});
+    if (!inserted) {
+        throw std::runtime_error("duplicate io_uring request id");
+    }
+    auto& state = iter->second;
+    state.expected_size = size;
+    state.ok = true;
+
+    size_t prepared = 0;
```
liliu-z, line 191: In `Submit()` (lines 110-132), when the SQ is full and `io_uring_submit` is called to flush, a failure leaves already-prepped-but-unsubmitted SQEs in the ring. Unlike `ReadUringAsync` where the ring returns to a pool, here the same `IOCompletionReader` retains the ring, so the stale SQEs (referencing caller buffers that may be freed after the thrown exception) will be inadvertently submitted on the next `Submit()` call, causing use-after-free. Fix: same strategy as the `ReadUringAsync` stale-SQE fix—drain or reset the SQ on failure.

src/knowhere/io_completion_reader.cc
```diff
@@ -0,0 +1,345 @@
...
+        auto completion = ready_completions_.front();
+        ready_completions_.pop_front();
+        return completion;
+    }
+
+    while (true) {
+        io_uring_cqe* cqe = nullptr;
+        const auto ret = io_uring_wait_cqe(handle_.uring, &cqe);
+        if (ret < 0) {
+            if (-ret == EINTR) {
```
liliu-z: Three methods have unbounded EINTR retry loops: `WaitCompleted()` (line 183), `PollCompleted()` (line 222), and `DrainOutstandingNoThrow()` (line 297). The same PR's commit `46cf89c` added bounded EINTR retries to `io_reader.cc` via `ShouldRetryInterruptedSyscall`, but `io_completion_reader.cc` was not updated consistently. `PollCompleted()` is particularly concerning because `io_uring_peek_cqe` is non-blocking, so sustained EINTR turns into a CPU spin loop. Fix: apply the same `ShouldRetryInterruptedSyscall` bounded retry from `io_reader_internal.h` to all three sites.

src/knowhere/io_completion_reader.cc
```diff
@@ -0,0 +1,396 @@
...
+            DrainOutstanding(request_id);
+        } catch (...) {
+            pending_requests_.erase(request_id);
+        }
+        RemoveReadyCompletion(request_id);
+    } else {
+        pending_requests_.erase(request_id);
+    }
+
+    if (prepared > submitted) {
```
liliu-z, line 533: `CleanupFailedSubmit` calls `ResetHandleUring()` (which does `io_uring_queue_exit` + `io_uring_queue_init`) when a single submission fails, but the same `IOCompletionReader` may still have other request IDs in `pending_requests_`. Resetting the ring discards CQEs for those older requests, so subsequent `DrainOutstanding` / `WaitCompleted` (line 157) blocks forever waiting for completions that will never arrive. The move-assignment path at line 55/61 is also affected because it calls the throwing `DrainOutstanding()`, making the hang reachable from `operator=` as well.

src/knowhere/io_completion_reader.cc
```diff
@@ -0,0 +1,396 @@
...
+    }
+    return std::nullopt;
+#else
+    (void)cqe;
+    return std::nullopt;
+#endif
+}
+
+void
+IOCompletionReader::DrainOutstandingNoThrow() noexcept {
```
liliu-z, line 419: `DrainOutstandingNoThrow` breaks out of the drain loop on `ProcessCqe` error (line 315) or after repeated EINTR/wait failure, then clears local bookkeeping (lines 321-322). `ReleaseHandle()` (line 391) subsequently pushes the ring back to the pool. If CQEs from abandoned requests are still pending in the kernel, the next consumer of that ring will observe foreign completions.

src/knowhere/uring_context_pool.cc
```diff
@@ -0,0 +1,201 @@
...
+
+UringContextPool::UringContextPool(size_t num_ctx, size_t max_entries) : num_ctx_(num_ctx), max_entries_(max_entries) {
+    ring_bak_.reserve(num_ctx_);
+
+    for (size_t i = 0; i < num_ctx_; ++i) {
+        auto* ring = new io_uring();
+        std::memset(ring, 0, sizeof(io_uring));
+        int ret = io_uring_queue_init(static_cast<unsigned>(max_entries_), ring, 0);
+        if (ret < 0) {
+            LOG_ERROR("io_uring_queue_init failed with ret=%d, errno=%d: %s", ret, -ret, ::strerror(-ret));
```
liliu-z: All 19 LOG calls across `uring_context_pool.cc`, `aio_context_pool.cc`, `io_context_pool.cc`, and `uring_context_pool.h` use printf-style placeholders (`%zu`, `%d`, `%s`, `%p`), but `LOG_ERROR`/`LOG_WARN` (defined at `include/log/Log.h:71`) route through `fmt::format(fmt::runtime(...))`. This produces literal `%zu` tokens in log output instead of actual values, defeating diagnostic logging on the very failure paths this PR introduces.

include/knowhere/io_completion_reader.h
```diff
@@ -0,0 +1,86 @@
...
+class IOCompletionReader {
+ public:
+    using RequestId = uint64_t;
+
+    struct Completion {
+        RequestId request_id = 0;
+        bool ok = false;
+    };
+
+    IOCompletionReader(int fd, std::shared_ptr<IOContextPool> io_pool = IOContextPool::GetGlobal());
```
liliu-z: The `IOCompletionReader` constructor's default argument `IOContextPool::GetGlobal()` can auto-initialize the global pool as a side effect of merely constructing a reader, which may surprise callers who expect explicit initialization.

src/knowhere/io_completion_reader.cc
```diff
@@ -0,0 +1,416 @@
...
+IOCompletionReader::DrainOutstanding(RequestId request_id) {
+#ifdef WITH_IO_URING
+    while (pending_requests_.find(request_id) != pending_requests_.end()) {
+        WaitCompleted();
+    }
+#endif
+}
+
+void
+IOCompletionReader::CleanupFailedSubmit(RequestId request_id, size_t prepared, size_t submitted) {
```
liliu-z, line 531: On the `prepared > submitted` path (L359–363), `CleanupFailedSubmit()` clears `pending_requests_` and `ready_completions_` for the entire reader, not just the failing request. If request A was already successfully submitted and request B's submit fails, request A's completions are silently lost. The `catch(...)` path at L370–374 has the same problem. This is a real multi-request correctness bug since `IOCompletionReader` explicitly supports multiple outstanding request IDs.

liliu-z:
**.github/workflows/ut.yaml:36** The CI workflow does not install `liburing-dev`, so the primary new io_uring code path introduced by this PR is compiled out via preprocessor guards and never exercised in CI. The two correctness bugs (stale SQEs, CleanupFailedSubmit) are completely untested in automated checks.

src/knowhere/io_completion_reader.cc
```diff
@@ -0,0 +1,416 @@
...
+    auto request_id = next_request_id_++;
+    auto [iter, inserted] = pending_requests_.emplace(request_id, RequestState{});
+    if (!inserted) {
+        throw std::runtime_error("duplicate io_uring request id");
+    }
+    auto& state = iter->second;
+    state.expected_size = size;
+    state.ok = true;
+
+    size_t prepared = 0;
```
liliu-z, line 191: At L117–150, `Submit()` never drains completion queue entries between submissions. For batches exceeding the CQ ring size (~256), this can cause stalls on 5.5+ kernels (which handle overflow via `IORING_FEAT_NODROP` backpressure rather than dropping). No batch size validation is performed.

test/IOContextPoolTest.cpp
```diff
@@ -0,0 +1,540 @@
...
+    ASSERT_TRUE(fut.get());
+    ASSERT_EQ(std::memcmp(first.get(), content.get(), kIOReaderTestBlockSize), 0);
+    ASSERT_EQ(std::memcmp(second.get(), content.get() + kIOReaderTestBlockSize, kIOReaderTestBlockSize), 0);
+
+    ::close(fd);
+    ::unlink(path);
+}
+
+#ifdef WITH_IO_URING
+TEST_F(IOContextPoolTestFixture, CompletionReaderReturnsSubmittedRequestIds) {
```
liliu-z, line 1085: The new test coverage at L475+ exercises only happy-path scenarios for `IOCompletionReader`. None of the new tests exercise the submit-failure or reset paths where the two main correctness bugs (stale SQEs, CleanupFailedSubmit orphaning requests) reside.

src/knowhere/uring_context_pool.cc
```diff
@@ -0,0 +1,199 @@
...
+}
+
+bool
+UringContextPool::ResetCheckedOut(struct io_uring* ring) {
+    if (ring == nullptr) {
+        LOG_WARN("UringContextPool reset gets null ring");
+        return false;
+    }
+
+    std::scoped_lock lk(ring_mtx_);
```
liliu-z: `ResetCheckedOut()` (lines 81-96) holds `ring_mtx_` while calling `io_uring_queue_exit()` and `io_uring_queue_init()`, which are blocking kernel syscalls. This serializes the entire pool — all `pop()` and `push()` operations block until the syscalls complete, creating a latency spike for all IO threads.

src/knowhere/io_context_pool.cc
```diff
@@ -0,0 +1,258 @@
...
+        LOG_INFO("Global IOContextPool initialized with backend aio");
+        return true;
+    }
+#endif
+
+    LOG_ERROR("Failed to initialize IOContextPool with any backend");
+    return false;
+}
+
+std::shared_ptr<IOContextPool>
```
liliu-z, line 195: `GetGlobal()` (lines 109-125) now auto-initializes the pool if not yet created, silently choosing the IO backend (io_uring vs AIO) based on runtime detection with default parameters. Unlike the legacy lazy init, backend selection is now part of auto-init, making silent defaults more consequential — callers may unknowingly get a different backend than intended.

include/knowhere/io_completion_reader.h
```diff
@@ -0,0 +1,98 @@
...
+    operator=(const IOCompletionReader&) = delete;
+
+    IOCompletionReader(IOCompletionReader&& other) noexcept;
+    IOCompletionReader&
+    operator=(IOCompletionReader&& other);
+
+    ~IOCompletionReader();
+
+    RequestId
+    Submit(std::span<std::byte* const> buffers, size_t size, std::span<const size_t> offsets);
```
liliu-z: The new public header includes `<span>` (line 8) and uses `std::span` in the API (line 38), requiring C++20. However, `CMakeLists.txt` does not add `cxx_std_20` as a public compile feature on the exported target. Downstream consumers that build with C++17 will fail to compile. Practical risk is low since known downstreams already use C++20.

test/IOContextPoolTest.cpp
```diff
@@ -0,0 +1,646 @@
...
+    EXPECT_THROW(reader.Submit(std::span<std::byte* const>(buffers.data(), buffers.size()), kIOReaderTestBlockSize,
+                               std::span<const size_t>(offsets.data(), offsets.size())),
+                 std::invalid_argument);
+
+    ASSERT_EQ(::close(fd), 0);
+    ::unlink(path);
+}
+
+#ifdef ENABLE_SYNCPOINT
+TEST_F(IOContextPoolTestFixture, CompletionReaderSubmitFailureKeepsExistingRequestObservable) {
```
liliu-z, line 1402: The existing syncpoint test (`CompletionReaderSubmitFailureKeepsExistingRequestObservable`, line 574) only validates `IOCompletionReader` recovery via `CleanupFailedSubmit`. There is no analogous test for `ReadUringAsync()`'s submit failure path — exactly where the dirty-ring bug (issue #1) lives. This test gap reduces confidence in the fix once applied.

src/knowhere/io_reader.cc
```diff
@@ -0,0 +1,583 @@
...
+            }
+            break;
+        }
+        if (completed == 0) {
+            if (++retry > kNumRetries) {
+                break;
+            }
+            continue;
+        }
+        for (size_t i = result.completed; i < result.completed + static_cast<size_t>(completed); ++i) {
```
liliu-z, line 213: The `retry` counter in `WaitAioBatch` accumulates across iterations that made forward progress. With `kNumRetries = 10` and `max_events = 128`, an AIO batch that the kernel returns piecewise will be cut off after 11 partial completions even when nothing is wrong. The io_uring sibling `WaitUringBatch` does not have this bug — it just increments `result.completed` per CQE with no retry counter on the progress path.

src/knowhere/io_reader.cc
```diff
@@ -0,0 +1,583 @@
...
+            if (-submitted == EINTR) {
+                if (!knowhere_internal::ShouldRetryInterruptedSyscall(retry, kNumRetries)) {
+                    break;
+                }
+                continue;
+            }
+            break;
+        }
+        if (submitted == 0) {
+            if (++retry > kNumRetries) {
```
liliu-z, line 161: Same progress-penalty bug as `WaitAioBatch`: `retry` accumulates even when `submitted > 0`, causing premature abort of legitimate partial submissions.

src/knowhere/io_reader.cc
```diff
@@ -0,0 +1,583 @@
...
+        }
+        active_ = false;
+        pool_->Reset(handle_);
+    }
+#endif
+
+ private:
+    std::shared_ptr<IOContextPool> pool_;
+    IOContextHandle handle_;
+    bool active_ = true;
```
liliu-z, line 82: The `WaitAioBatch` function takes `const std::vector<struct iocb>& cbs` but uses `const_cast<struct iocb*>(&cbs[i])` to insert pointers into an `unordered_set<struct iocb*>`. This strips const-correctness and risks undefined behavior if the underlying data is truly const.

src/knowhere/io_completion_reader.cc
```diff
@@ -0,0 +1,563 @@
...
+        if (ret < 0) {
+            if (-ret == EINTR) {
+                if (!knowhere_internal::ShouldRetryInterruptedSyscall(retry, kNumRetries)) {
+                    throw std::runtime_error("io_uring_wait_cqe interrupted too many times");
+                }
+                continue;
+            }
+            throw std::runtime_error("io_uring_wait_cqe failed");
+        }
+        if (cqe == nullptr) {
```
liliu-z, line 340: After `io_uring_wait_cqe` returns success (ret == 0), if `cqe` is `nullptr` the code does `continue` with no retry cap, no progress tracking, and no logging. If the kernel ever returns this combination, the worker thread hangs in an infinite loop. Compare `ProcessAvailableCompletions` at line 336 which correctly breaks on null cqe.

liliu-z:
**src/common/io_reader.cc:0** `~IOCompletionReader` calls `io_uring_queue_exit()`, which triggers kernel cancellation asynchronously via a workqueue (`io_ring_exit_work`). `close(ring_fd)` returns to userspace before in-flight DMA completes. If the caller then frees the target buffers, the kernel's pending writes corrupt the reallocated memory — silent data corruption, not a crash. The destructor must synchronously drain all outstanding CQEs (observing `-ECANCELED`) before returning, or the API contract must enforce that callers cannot free buffers until all completions are observed.

liliu-z:
**src/common/io_reader.cc:0** When `ReadAioAsync` or `ReadUringAsync` encounters a mid-batch failure (hardware error, short read, bad CQE), it calls `state.ResetRing()` or `guard.ResetAio()` (triggering `io_uring_queue_exit` / `io_destroy`), then returns `false` via the future. The caller sees failure and frees the buffer, but kernel DMA cancellation from `io_destroy`/`io_uring_queue_exit` is fully asynchronous — the block layer may still be writing into the now-freed memory. This occurs on *any* normal IO error return path, not just destruction.

CMakeLists.txt
```diff
@@ -115,10 +115,33 @@ else()
     list(APPEND COMMON_LINKER_LIBS ${LIBAIO_LIBRARY})
 endif()
 
+set(MILVUS_COMMON_WITH_IO_URING OFF)
+find_path(URING_INCLUDE_DIR liburing.h)
```
liliu-z, line 119: The C++20 standard is linked as `PUBLIC`, meaning every target that depends on this library is forced to compile with C++20. This is driven by `std::span` usage in headers. Downstream projects that cannot adopt C++20 will fail to build. The requirement should be `PRIVATE`, and `std::span`-using headers should be isolated or guarded.

liliu-z:
**src/common/io_reader.cc:0** When the AIO backend is used with `O_DIRECT` file descriptors, the kernel requires buffer addresses and sizes to be aligned (typically 512-byte or page-aligned). `IOReader::Read` and `IOReader::ReadAsync` do not validate alignment at the API boundary. Misaligned requests silently fail with `EINVAL` deep in the kernel, giving callers no actionable diagnostic.

liliu-z:
**src/common/aio_context_pool.h:151** The legacy `AioContextPool::Shutdown()` is public and called by tests; `UringContextPool` has no equivalent method. This creates an API asymmetry that forces callers to branch on backend type for lifecycle management.

liliu-z:
**src/common/io_context_pool.cc:0** `generation_` is set once during `InitGlobal()` and never mutated afterward. The condition `owner_generation_ != generation_` in `Release` is unreachable in current code — a handle's `owner_` shared_ptr keeps the pool alive, so generation always matches. The defensive `ClearNoRelease` branch would leak the context if it ever fired. This is a latent trap if `generation_` becomes mutable in the future.

src/knowhere/io_context_pool.cc
```diff
@@ -0,0 +1,416 @@
+#include "knowhere/io_context_pool.h"
```
liliu-z, line 1: Enabling `WITH_IO_URING` causes `GetGlobalAioPool()` to return `nullptr`, breaking downstream consumers (knowhere/cardinal) that rely on this API. This is an uncoordinated contract change that will cause crashes in callers that don't null-check the return value.

src/knowhere/io_completion_reader.cc
```diff
@@ -0,0 +1,561 @@
...
+    return *this;
+}
+
+IOCompletionReader::~IOCompletionReader() {
+    DrainOutstandingNoThrow();
+    ReleaseHandle();
+}
+
+IOCompletionReader::RequestId
+IOCompletionReader::Submit(IOCompletionReaderSpan<std::byte* const> buffers, size_t size,
```
liliu-z, line 142: IOReader::ReadAsync validates O_DIRECT alignment via `ValidateDirectIoAlignment` (io_reader.cc:607), but IOCompletionReader::Submit (lines 102-187) has no such check. With an O_DIRECT fd, unaligned buffers will produce a confusing kernel-side EINVAL instead of a clear library-level error. This creates inconsistent error semantics between the two reader classes in the same library.

src/knowhere/io_context_pool.cc
```diff
@@ -0,0 +1,416 @@
...
+}
+
+#ifdef WITH_IO_URING
+bool
+IOContextPool::TryInitUring(const IOContextPoolConfig& cfg, const std::shared_ptr<IOContextPool>& io_pool) {
+    if (!UringContextPool::InitGlobalUringPoolWithValidation(cfg.num_ctx, cfg.max_events)) {
+        return false;
+    }
+
+    auto pool = UringContextPool::GetGlobalUringPoolDirect();
```
liliu-z, line 102: IOContextPool has `Push(IOContextHandle&)` lvalue overloads that silently move the handle's ownership. A caller passing an lvalue won't realize ownership transferred. This is a classic implicit-move footgun that can lead to double-free or use-after-move bugs.

test/IOContextPoolTest.cpp
```diff
@@ -0,0 +1,1582 @@
...
+    sync_point->ClearAllCallBacks();
+    sync_point->ClearTrace();
+
+    auto pool = IOContextPool::GetGlobal();
+    ASSERT_NE(pool, nullptr);
+    ASSERT_EQ(pool->Backend(), IOBackend::AIO);
+}
+#endif
+#endif
+
```
liliu-z, line 109: The test uses a 100ms sleep as a synchronization mechanism. Under ASan or loaded CI environments, this timing window is unreliable and will produce flaky failures. The test needs proper synchronization (e.g., condition variable or future) instead of a sleep.

src/knowhere/io_context_pool.cc
```diff
@@ -0,0 +1,416 @@
...
+void
+IOContextHandle::ReleaseNoThrow() noexcept {
+    if (!HasContext()) {
+        ClearNoRelease();
+        return;
+    }
+
+    auto owner = owner_;
+    if (owner == nullptr) {
+        LOG_WARN("IOContextHandle drops context without owner for backend {}", static_cast<int>(backend));
```
liliu-z, line 79: AIO's `ResetCheckedOut` and uring's reset follow different patterns for returning the new context to the pool. This asymmetry makes the code harder to reason about and creates risk of resource leaks if a maintainer assumes symmetric behavior.

src/knowhere/io_reader.cc
```diff
@@ -0,0 +1,641 @@
...
+
+    auto ctx = handle.aio;
+    IOContextHandleGuard guard(pool, std::move(handle));
+    const size_t max_batch = pool->MaxEventsPerCtx();
+    if (max_batch == 0) {
+        return MakeReadyFuture(false);
+    }
+
+    const size_t first_batch = std::min(max_batch, buffers.size());
+    std::vector<struct iocb> first_cbs;
```
liliu-z: After `io_submit`, the kernel holds pointers to iocb addresses inside `first_cbs`. The `std::move(first_cbs)` into `AioReadState::first_cbs_` is safe because vector move is a pointer swap with the default allocator. However, this invariant is fragile — switching to SmallVector or a custom allocator would silently break it by relocating the buffer and corrupting kernel-visible addresses.

src/knowhere/io_completion_reader.cc
```diff
@@ -0,0 +1,561 @@
...
+    other.next_request_id_ = 1;
+}
+
+IOCompletionReader&
+IOCompletionReader::operator=(IOCompletionReader&& other) {
+    if (this == &other) {
+        return *this;
+    }
+
+    DrainOutstandingNoThrow();
```
liliu-z, line 120: `IOCompletionReader::operator=(&&)` is not `noexcept` even though the move constructor is. This asymmetry prevents optimal container behavior (e.g., `std::vector` reallocation will copy instead of move).

src/knowhere/io_reader.cc
```diff
@@ -0,0 +1,641 @@
...
+    int forced_ret = 0;
+    TEST_SYNC_POINT_CALLBACK("IOReader::SubmitUring:Before", &forced_ret);
+    if (forced_ret != 0) {
+        return forced_ret;
+    }
+#endif
+    return io_uring_submit(ring);
+}
+
+size_t
```
liliu-z, line 368: PrepareUringBatch doesn't apply an explicit `std::min(buffers.size() - start, max_events_per_ctx)` cap like ReadAioAsync does. While `io_uring_get_sqe` returning null provides a natural cap via the SQ ring size, the asymmetry with the AIO path creates a readability and maintenance concern.

yhmo:
@Mergifyio refresh

yhmo (commented):
@dcoapp recheck

PR #81, opened 2026-04-24
Manage openblas by conan

yhmo:


conanfile.py
```diff
@@ -60,6 +62,8 @@ def requirements(self):
         self.requires("nlohmann_json/3.11.3#ffb9e9236619f1c883e36662f944345d", force=True)
         if self.settings.os != "Macos":
             self.requires("libunwind/1.8.1#748a981ace010b80163a08867b732e71")
+            # openblas is only used on Linux (thread_pool.cc is guarded by __linux__)
+            self.requires("openblas/0.3.27")
```
alexanderguzhva, line 66: @yhmo this is a Problem.
We must ALWAYS use the most recent openblas and/or even maintain our own conan recipe for the the most recent openblas version (which is 0.3.32 as of today). The reason is that OpenBLAS may contain bugs, and previously I've seen a problem for PCA related to OpenBLAS
An easy temporary solution would be to use 0.3.30, which is the most available one

foxspy (approved):
/lgtm

foxspy (approved):

PR #83, opened 2026-04-28
Update Makefile and README

yhmo:
If your Conan profile defaults to a lower C++ standard(cppstd=17 or cppstd=gnu17) , then you will get compile errors since it requires minimum_cpp_standard=20. 
Now we add an option "CONAN_EXTRA" to pass the cppstd to conan.
And the "NPROC" is also passed to conan so that we can control the cpu usage for conan during compiling.
This pr also updated the README


foxspy (approved):
/lgtm

PR #84, opened 2026-04-29
Add macOS unit test workflow

yhmo:


.github/workflows/ut.yaml
```diff
@@ -53,3 +57,41 @@ jobs:
...
+      - name: Install Dependency
+        run: |
+          python3 -m venv "$RUNNER_TEMP/conan-venv"
+          "$RUNNER_TEMP/conan-venv/bin/python" -m pip install conan==2.25.1 "cmake<4"
+          echo "$RUNNER_TEMP/conan-venv/bin" >> $GITHUB_PATH
+      - name: Restore Conan Packages
+        uses: actions/cache@v4
+        with:
+          path: ~/.conan2
+          key: milvus-common-macos-15-${{ hashFiles('conanfile.py')}}
```
liliu-z, line 81: Runner is `macos-15` (line 63) but cache key/restore-key/save key at lines 81, 82, and 94 all use `macos-14`. After the first run populates the cache, subsequent runs will hit it (so it's not a persistent miss), but the key is semantically wrong. If a real `macos-14` job is added later, the two runner images will share the same cache namespace, risking silent ABI-incompatible cache pollution. Fix: replace `macos-14` with `macos-15` in all three locations.

.github/workflows/ut.yaml
```diff
@@ -16,11 +16,15 @@ on:
       - 'CMakeLists.txt'
       - '.github/workflows/ut.yaml'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
-  ut:
+  ut-linux:
```
liliu-z, line 24: The job ID changed from `ut` to `ut-linux`. If the repository's branch protection settings require the old `ut` status check, PRs will either hang waiting for a check that never arrives or lose their merge gate. This cannot be verified from code alone — confirm in the GitHub repo settings before merging and update any required status checks accordingly.

conanfile.py
```diff
@@ -57,19 +60,23 @@ def requirements(self):
...
             self.requires("libunwind/1.8.1#748a981ace010b80163a08867b732e71")
+        if self.settings.os == "Linux":
             # openblas is only used on Linux (thread_pool.cc is guarded by __linux__)
             self.requires("openblas/0.3.30")
 
     @property
     def _minimum_cpp_standard(self):
         return 20
 
+    def validate(self):
```
liliu-z, line 77: The new `validate()` calls `check_min_cppstd(self, 20)` which requires `compiler.cppstd` to be explicitly defined in the Conan profile (confirmed via Conan 2.25.1 source: `None` raises `ConanInvalidConfiguration`). Previously, consumers could omit `cppstd` and rely on CMake's `set(CMAKE_CXX_STANDARD 20)`. This is a correct fail-early improvement, but it tightens the interface — consumers must now pass `-s compiler.cppstd=20` explicitly. Consider documenting this in the PR description or README.

Makefile
```diff
@@ -42,6 +44,7 @@ configure: conan
 build: configure
 	cmake --build $(BUILD_DIR) -j$(NPROC)
 
+test: CONAN_EXTRA += -o \&:with_ut=True
```
liliu-z: GNU Make target-specific `+=` cannot append to command-line variables (they have highest priority). Running `make test CONAN_EXTRA="-o &:with_asan=True"` silently drops `with_ut=True`, so Conan won't install gtest and `find_package(GTest REQUIRED)` in CMakeLists.txt:147 will fail at configure time. Locally reproduced with a minimal Makefile. The same applies to line 48 (`CMAKE_EXTRA`). Fix: use `test: override CONAN_EXTRA += -o \&:with_ut=True` (and likewise for `CMAKE_EXTRA`).

conanfile.py
```diff
@@ -48,6 +47,10 @@ class MilvusCommonConan(ConanFile):
         "with_asan": False,
     }
 
+    def build_requirements(self):
```
liliu-z, line 50: gtest is now only installed when `with_ut=True` (via `test_requires`), but the current README (line 48) still shows `make ... CMAKE_EXTRA="-DWITH_COMMON_UT=ON"` using the default `all` target. The default target does not trigger the `test:` target-specific variable that injects `with_ut=True`, so Conan won't install gtest while CMakeLists.txt:147 still calls `find_package(GTest REQUIRED)` — cmake configure fails. README is not in this PR's diff, but the interaction is introduced by this PR. Fix: update the README example to use `make test`, or replace the `CMAKE_EXTRA="-DWITH_COMMON_UT=ON"` example with a non-gtest flag.

.github/workflows/ut.yaml
```diff
@@ -16,11 +16,15 @@ on:
...
     name: UT on ubuntu-22.04
     runs-on: ubuntu-22.04
-    timeout-minutes: 240
+    timeout-minutes: 60
```
liliu-z, line 27: Linux CI's `CONAN_EXTRA` explicitly passes `-o &:with_ut=True` and `-s compiler.cppstd=20`, but both are now automatically provided by the Makefile: `with_ut` via `test: override CONAN_EXTRA += -o &:with_ut=True` (L47) and `cppstd` via `-s compiler.cppstd=$(CONAN_CPPSTD)` (L32, default `20`). Conan receives duplicate values and takes the last one, so behavior is correct, but the redundancy creates a maintenance risk — if the Makefile default changes without updating this CI line (or vice versa), a silent inconsistency could arise.

foxspy (approved):
/lgtm

PR #86, opened 2026-05-06
refactor: reduce loading overhead tracker lock contention

sunby:
## Summary

This PR reduces lock contention in `LoadingOverheadTracker` by splitting the previous single global mutex into:

- a `std::shared_mutex` for group/handle map access
- a per-group `GroupState` mutex for reservation state updates

The hot-path operations (`Reserve`, `Release`, `HasFiniteUpperBound`, and `GetUpperBound`) now take a shared map lock, copy the target group state pointer, and only serialize updates within that group. This allows independent groups to reserve/release loading overhead without blocking each other on one global tracker mutex.

Registration and cleanup behavior is preserved:

- existing groups still reuse the same handle and increment ref counts
- finite upper bounds still take the max per dimension when re-registered
- unlimited upper bounds can still be replaced by the registered bound
- `Unregister` still removes the group at ref count 0 and logs residual reservations before cleanup

## Testing

- `pre-commit` passed
- `UT on ubuntu-22.04` passed
- `UT on macos-15` passed

### DIFF
--- a/include/cachinglayer/LoadingOverheadTracker.h
+++ b/include/cachinglayer/LoadingOverheadTracker.h
@@ -134,8 +134,8 @@ class LoadingOverheadTracker {
                           group, handle, ref_count);
                 break;
             case LogEvent::kNewGroup:
-                LOG_INFO("[MCL] LoadingOverheadTracker registered group '{}' (handle {}, refs=1): UB={}", group,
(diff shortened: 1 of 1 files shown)


yhmo:
@Mergifyio refresh

yhmo (commented):
@dcoapp recheck

PR #87, opened 2026-05-18
Tune search OMP setter for direct HNSW

CLiqing:
This updates `ThreadPool::ScopedSearchOmpSetter` so callers can choose whether to also change OpenBLAS thread settings.

For direct HNSW search execution, the caller may need to clamp OpenMP search parallelism without repeatedly changing the global OpenBLAS thread count. The default behavior is preserved by keeping `set_blas_threads=true`.

Changes:
- Store and restore the actual previous OpenMP max thread count.
- Add `set_blas_threads` parameter, defaulting to `true`.
- Only call `openblas_set_num_threads` when `set_blas_threads` is enabled.

Validation:
- Built and tested in the Milvus HNSW search-pool bypass experiment.
- Used by the follow-up Knowhere direct HNSW path to avoid unnecessary OpenBLAS global thread updates.

### DIFF
--- a/include/knowhere/thread_pool.h
+++ b/include/knowhere/thread_pool.h
@@ -223,10 +223,11 @@ class ThreadPool {
     class ScopedSearchOmpSetter {
         int omp_before;
 #ifdef OPENBLAS_OS_LINUX
+        bool set_blas_threads;
         int blas_thread_before;
 #endif
      public:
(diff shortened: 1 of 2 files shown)


CLiqing:
Closing this for now. We will keep the direct HNSW experiment independent from milvus-common OMP setter changes and revisit only if needed.

PR #93, opened 2026-06-11
enhance: Add trace context snapshot to OpContext

congqixia:
Store an owned trace snapshot in OpContext so cache-layer code can propagate tracing metadata without depending on borrowed TraceContext buffers. Expose helpers to set trace data and create temporary TraceContext views for the existing tracer API.

Keep async warmup isolated from caller tracing by constructing a fresh OpContext with only the cancellation token. Add tracer and cache slot tests to lock in the snapshot lifetime and warmup propagation semantics.


include/common/Tracer.h
```diff
@@ -39,6 +41,46 @@ struct TraceContext {
...
+    OwnedTraceContext() = default;
+
+    explicit OwnedTraceContext(const TraceContext& ctx) {
+        if (ctx.traceID == nullptr || ctx.spanID == nullptr) {
+            return;
+        }
+        std::copy_n(ctx.traceID, trace_id.size(), trace_id.begin());
+        std::copy_n(ctx.spanID, span_id.size(), span_id.begin());
+        trace_flags = ctx.traceFlags;
+        has_value = true;
```
liliu-z, line 71: The OwnedTraceContext(const TraceContext&) constructor sets has_value=true whenever the source pointers are non-null, but never checks whether the ID bytes are all-zero. The codebase treats an all-zero ID as 'no trace' (isEmptyID/EmptyTraceID in Tracer.cpp), so a non-null but all-zero traceID is stored here as a present trace, making HasValue()/MakeTraceContextView() report a trace the rest of the code considers empty. StartSpan re-guards against empty IDs so span creation itself stays safe, but presence checks keyed off the snapshot are inconsistent with the convention. Normalize through isEmptyID before setting has_value.

include/common/OpContext.h
```diff
@@ -38,6 +40,29 @@ struct OpContext {
     // Maps to proto::common::LoadPriority (milvus-proto): HIGH = 0, LOW = 1.
     std::optional<int32_t> runtime_load_priority;
 
+    std::optional<tracer::OwnedTraceContext> trace_context;
+
+    void
+    SetTraceContext(const tracer::TraceContext& ctx) {
+        trace_context.emplace(ctx);
```
liliu-z: SetTraceContext is not safe when ctx aliases this OpContext's own stored snapshot (e.g. op_ctx.SetTraceContext(*op_ctx.MakeTraceContextView())): trace_context.emplace(ctx) destroys the existing OwnedTraceContext and reconstructs one in the same storage, whose NSDMI trace_id{}/span_id{} zero-fills the arrays before the constructor's copy_n reads from the now-zeroed, aliased source — silently overwriting the trace with zeros. No in-repo caller does this today so it is latent, but since SetTraceContext is public library API, build the snapshot in an independent temporary and move it in to be robust against aliased input.

include/common/OpContext.h
```diff
@@ -16,6 +16,8 @@
 #include <cstdint>
 #include <optional>
 
+#include "common/Tracer.h"
```
liliu-z, line 19: This PR newly adds `#include "common/Tracer.h"` to the foundational OpContext.h, which transitively pulls in opentelemetry/trace/provider.h. OpContext.h is included by the caching-layer headers (DList.h, Translator.h, CacheSlot.h), none of which depended on OpenTelemetry before this change, so the whole caching layer (and downstream includers) now compiles against the OTel provider header. OpContext.h genuinely needs the OwnedTraceContext/TraceContext definitions so the include itself is required; the lighter fix is to have Tracer.h pull trace_id.h/span_id.h (plus forward declarations) instead of provider.h, or to relocate OwnedTraceContext into a lighter header.

include/common/Tracer.h
```diff
@@ -39,6 +47,51 @@ struct TraceContext {
...
+    HasValue() const {
+        return has_value;
+    }
+
+    [[nodiscard]] TraceContext
+    AsTraceContext() const {
+        if (!has_value) {
+            return {};
+        }
+        return TraceContext{trace_id.data(), span_id.data(), trace_flags};
```
liliu-z, line 84: AsTraceContext() returns a TraceContext whose pointers alias this object's internal arrays, so the view dangles if it is called on a temporary or kept after the backing store goes away — including the realistic case where OpContext::MakeTraceContextView() hands a view to cache-layer code that then outlives the OpContext or calls ClearTraceContext()/SetTraceContext(). No current call site misuses it, so this is hardening rather than a live bug: consider ref-qualifying the method `const &` (and `= delete` the `const &&` overload) and documenting that the view must be consumed immediately and not retained across Set/Clear.

include/common/Tracer.h
```diff
@@ -11,12 +11,20 @@
...
 #include <map>
 #include <memory>
 #include <string>
 
 #include "common/TracerBase.h"
-#include "opentelemetry/trace/provider.h"
+#include "opentelemetry/trace/span_id.h"
+#include "opentelemetry/trace/trace_id.h"
+
+namespace opentelemetry::trace {
```
liliu-z, line 24: Public Tracer.h dropped `#include "opentelemetry/trace/provider.h"` and now only forward-declares Span/Tracer. The forward-declaration is itself correct (span_id.h/trace_id.h above it establish the inline v1::trace namespace, so the decls land in the right namespace and the repo still builds), but downstream consumers that previously got the complete Span/Tracer definitions transitively through Tracer.h and call methods on them (e.g. span->End()) will now hit incomplete-type errors and must add a direct otel include. Validate against a Milvus build; the fix is a one-line include downstream (or restore the provider.h include here).

liliu-z (commented):
/lgtm
/approve

PR #94, opened 2026-06-15
feat: add helpers for creating feature usage counter

zhengbuqian:
issue: https://github.com/milvus-io/milvus/issues/51149

### DIFF
--- a/include/common/FeatureReport.h
+++ b/include/common/FeatureReport.h
@@ -1,0 +1,79 @@
+// Copyright (C) 2019-2020 Zilliz. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software distributed under the License
+// is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+// or implied. See the License for the specific language governing permissions and limitations under the License.
+
+#pragma once
+
+#include <prometheus/counter.h>
+
+#include <atomic>
+#include <chrono>
+#include <string>
+#include <string_view>
+
+namespace milvus::monitor::feature_report {
+
+constexpr std::string_view kHybridSearch = "hybrid_search";
+constexpr std::string_view kPartitionKey = "partition_key";
+constexpr std::string_view kDynamicField = "dynamic_field";
+constexpr std::string_view kBM25Function = "bm25_function";
+constexpr std::string_view kResourceGroup = "resource_group";
+constexpr std::string_view kBulkImport = "bulk_import";
+
+class alignas(64) FeatureReporter {
+ public:
+    // Prefer the predeclared static reporters below on hot paths.
+    explicit FeatureReporter(std::string_view name);
+
+    FeatureReporter(const FeatureReporter&) = delete;
+    FeatureReporter&
+    operator=(const FeatureReporter&) = delete;
+
+    std::string_view
+    Name() const noexcept;
+
+    bool
+    Record();
+
+    bool
+    RecordAtForTest(std::chrono::steady_clock::time_point now);
+
+    void
+    ResetForTest();
+
+ private:
+    bool
+    recordAt(std::chrono::steady_clock::time_point now);
+
+ private:
+    std::atomic<int64_t> next_allowed_nanos_{0};
+    std::string name_;
+    prometheus::Counter& counter_;
+};
+
+FeatureReporter&
+HybridSearch();
+
+FeatureReporter&
+PartitionKey();
+
+FeatureReporter&
+DynamicField();
+
+FeatureReporter&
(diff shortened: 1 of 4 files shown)


include/common/FeatureReport.h
```diff
@@ -0,0 +1,70 @@
...
+    FEATURE(PartitionKey, "partition_key")   \
+    FEATURE(DynamicField, "dynamic_field")   \
+    FEATURE(BM25Function, "bm25_function")   \
+    FEATURE(ResourceGroup, "resource_group") \
+    FEATURE(BulkImport, "bulk_import")
+
+class alignas(64) FeatureReporter {
+ public:
+    // Prefer the predeclared static reporters below on hot paths.
+    explicit FeatureReporter(std::string_view name);
```
liliu-z: FeatureReporter exposes a public constructor and prometheus-cpp's Family::Add coalesces identical label sets, so constructing a second reporter with an existing feature name (e.g. FeatureReporter{kHybridSearch} alongside the HybridSearch() singleton) hands back the same shared Counter while giving the new instance its own next_allowed_nanos_ throttle. The ~once-per-hour invariant then holds per instance rather than per counter, so a feature can be recorded up to once per hour per duplicate reporter. Funnel construction through the singletons (e.g. make the constructor private) so each counter has exactly one throttle.

test/FeatureReportTest.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include "common/FeatureReport.h"
+#include "common/PrometheusClient.h"
+
+namespace milvus::monitor::feature_report {
+namespace {
+
+bool
+MetricsContain(const std::string& metrics, const std::string& feature, const std::string& value) {
+    const auto expected = "milvus_feature_report_total{feature=\"" + feature + "\",source=\"cpp\"} " + value;
+    return metrics.find(expected) != std::string::npos;
```
liliu-z: MetricsContain matches with an unanchored find(), and `expected` ends right after the counter value with no trailing newline. The feature name is actually bounded by the surrounding `",source="cpp"}`, so the reviewer's name-substring example does not apply here; the unbounded part is the value, so asserting a value of "2" would also pass against an actual value of "20" or "25". Append "\n" to `expected` (or match the whole line) so the count is checked exactly.

include/common/FeatureReport.h
```diff
@@ -0,0 +1,75 @@
...
+    std::string_view
+    Name() const noexcept;
+
+    bool
+    Record();
+
+    bool
+    RecordAtForTest(std::chrono::steady_clock::time_point now);
+
+    void
```
liliu-z: `RecordAtForTest()` and `ResetForTest()` are declared under `public:` (lines 43-47), exposing test-only seams — clearing the throttle gate and injecting arbitrary timestamps — on the production reporter's API. Since the class already `friend`s its factory functions, a friend-based test seam would keep these off the public surface. No in-repo production caller exists today, so this is an encapsulation concern rather than a live bug.

src/common/FeatureReport.cpp
```diff
@@ -0,0 +1,100 @@
...
+namespace {
+
+constexpr std::string_view kSourceCpp = "cpp";
+constexpr auto kReportInterval = std::chrono::hours(1);
+
+prometheus::Family<prometheus::Counter>&
+FeatureReportFamily() {
+    static auto& family = prometheus::BuildCounter()
+                              .Name("milvus_feature_report_total")
+                              .Help("Count of throttled feature reports.")
```
liliu-z: The counter increments only on the path that passes the throttle, so it measures emitted reports, not suppressed ones. `Help("Count of throttled feature reports.")` reads naturally as a count of dropped/rate-limited reports — the inverse of what is measured. Reword to something like "Count of feature reports emitted."

test/FeatureReportTest.cpp
```diff
@@ -0,0 +1,75 @@
...
+TEST(FeatureReportTest, Throttle) {
+    auto& reporter = HybridSearch();
+    reporter.ResetForTest();
+    const auto now = std::chrono::steady_clock::time_point(std::chrono::seconds(100));
+
+    ASSERT_TRUE(reporter.RecordAtForTest(now));
+    ASSERT_FALSE(reporter.RecordAtForTest(now + std::chrono::minutes(1)));
+    ASSERT_TRUE(reporter.RecordAtForTest(now + std::chrono::hours(1)));
+
+    ASSERT_TRUE(MetricsContain(getPrometheusClient().GetMetrics(), std::string(reporter.Name()), "2"));
```
liliu-z: The assertion hard-codes the absolute cumulative counter value — `MetricsContain(..., reporter.Name(), "2")` at line 41 (and "1" at line 65) — but the Prometheus counter is process-global and monotonic and `ResetForTest()` clears only the throttle gate, not the counter. Under `--gtest_repeat=2`, shuffled re-runs, or any later test that also records `HybridSearch`/`PartitionKey`, the value climbs to 4/2 and the assertion fails on otherwise-correct code. Read the counter before and after and assert the delta (+2/+1) instead of an absolute string; the unanchored `find` in `MetricsContain` coupling to integer formatting is a secondary concern.

PR #95, opened 2026-06-17
Add mergify

yhmo:


.github/mergify.yml
```diff
@@ -0,0 +1,67 @@
...
+    actions:
+      label:
+        remove:
+          - ci-passed
+
+  - name: Add ci-passed when no code changes
+    conditions:
+      - or: *BRANCHES
+      - "check-success=pre-commit"
+      - "-files~=^(?!(doc/|docs/|.*\\.md$|\\.gitignore$|LICENSE$|OWNERS$)).+"
```
yhmo: Medium ---- This "no code changes" guard is narrower than the repo’s UT path filter. `ut.yaml` only runs for `include/**`, `src/**`, `test/**`, `conanfile.py`, `Makefile`, `CMakeLists.txt`, and `.github/workflows/ut.yaml`, so a PR like this one that only changes `.github/mergify.yml` gets `pre-commit` but never produces the two `UT on ...` checks. At the same time, `.github/mergify.yml` does not match this exemption list, so it also misses the fallback rule here. The result is that config-only PRs like this never receive `ci-passed`. Could we either expand this allowlist to cover non-code infra files (for example `.github/**`) or relax the code-changed rule to account for paths that do not trigger UT at all?

PR #96, opened 2026-06-17
Add OpContext trace span helpers

jamesgao-jpg:
## Summary

- add `milvus::OpContext::trace_span` as the trace parent carrier for downstream search paths
- add `milvus::OpContext::GetTraceSpan()` accessors for trace span reads
- add shared trace helper APIs in `common/Tracer.h` for Cardinal and Knowhere
- remove the separate `common/Trace.h` wrapper; callers should include `common/Tracer.h` directly
- rename the scoped span-slot restoration helper to `NestedSpanGuard`
- remove thin wrapper helpers `SetTemporaryOpContextTraceSpan` and `StartScopedSpan`; callers now construct `NestedSpanGuard` and `ScopedSpan` directly
- keep temporary span replacement scoped so the previous span is restored when the guard is destroyed

## Dependency context

This PR is consumed by downstream tracing PRs:
- Cardinal: https://github.com/zilliztech/cardinal/pull/865
- Knowhere: https://github.com/zilliztech/knowhere/pull/1684
- Milvus: https://github.com/milvus-io/milvus/pull/50606

The downstream PRs currently pin the published helper package `milvus-common/1.0.0-45bff01@milvus/dev#97948822f1dacdeafed7e8d5251df493`, produced from pre-rebase commit `45bff01293e050be7af2c9c1e5b70c49da8940ef`. The branch has since been rebased on current `master`; source content is carried by the current PR head.

## Verification

- `git diff --check`
- No local build, test, Conan build/install, or Milvus validation was run after the latest rebase.


include/common/Tracer.h
```diff
@@ -145,6 +147,115 @@ GetSpanIDAsHexStr(const TraceContext* ctx);
...
+    NestedSpanGuard(SpanPtr& span_slot, const SpanPtr& current) : NestedSpanGuard(&span_slot, current) {
+    }
+
+    NestedSpanGuard(NestedSpanGuard&& other) noexcept
+        : span_slot_(other.span_slot_), previous_(std::move(other.previous_)) {
+        other.span_slot_ = nullptr;
+    }
+
+    NestedSpanGuard&
+    operator=(NestedSpanGuard&& other) noexcept {
```
liliu-z: NestedSpanGuard is a save/restore guard over an externally-owned span slot, yet it is left move-assignable. Move-assigning one guard onto another that targets the same slot corrupts that slot: the assignment first restores this guard's saved previous_ into the slot (overwriting the current span the source just installed), then adopts the source's state, so the slot ends up pointing at an older, already-popped span instead of the current one. Nothing in the PR needs move-assignment; make NestedSpanGuard non-movable like std::lock_guard by deleting the move operations.

include/common/OpContext.h
```diff
@@ -65,6 +66,19 @@ struct OpContext {
         return trace_context->AsTraceContext();
     }
 
+    // Current trace span propagated across Milvus/Knowhere/Cardinal boundaries.
+    std::shared_ptr<tracer::trace::Span> trace_span = nullptr;
```
liliu-z, line 72: OpContext::trace_span is a plain non-atomic, non-thread_local shared_ptr that NestedSpanGuard mutates in place, but the field is explicitly meant to be propagated across Milvus/Knowhere/Cardinal boundaries into the multithreaded search path. The existing span infrastructure is deliberately thread_local to avoid locking, and an OpContext is already touched by multiple threads (its storage_usage counters are std::atomic); if two workers share one OpContext and each installs a guard on trace_span, the in-place writes race on a non-atomic shared_ptr. Document a single-threaded/per-worker ownership model for the slot, or make it per-thread.

include/common/Tracer.h
```diff
@@ -145,6 +147,115 @@ GetSpanIDAsHexStr(const TraceContext* ctx);
...
+        : span_slot_(span_slot), previous_(span_slot != nullptr ? *span_slot : SpanPtr{}) {
+        if (span_slot_ != nullptr) {
+            *span_slot_ = current;
+        }
+    }
+
+    NestedSpanGuard(SpanPtr& span_slot, const SpanPtr& current) : NestedSpanGuard(&span_slot, current) {
+    }
+
+    NestedSpanGuard(NestedSpanGuard&& other) noexcept
```
liliu-z: NestedSpanGuard's move constructor copies the raw span_slot_ pointer into the new guard (only nulling the source), so a moved-to guard can outlive the OpContext whose member the slot points into, after which the destructor's `*span_slot_ = previous_` writes through a dangling pointer. Under C++20 guaranteed copy elision the test's `auto g = NestedSpanGuard(...)` does not need this constructor and nothing returns/relocates a guard by value, so it is unused dead surface — delete it together with the move-assignment to make the guard non-movable and close the latent use-after-free.

jamesgao-jpg: Fixed in 40659ec239cfae17118c956ac4488a9d04576bb5 by deleting NestedSpanGuard move construction together with move assignment. The guard now stays bound to the slot it was constructed for and cannot be relocated with a raw span_slot_ pointer.

include/common/Tracer.h
jamesgao-jpg: Fixed in 40659ec239cfae17118c956ac4488a9d04576bb5 by deleting NestedSpanGuard move-assignment. NestedSpanGuard is now non-movable, so it cannot restore/adopt state across the same external span slot.

include/common/OpContext.h
jamesgao-jpg: Fixed in 75fd9691a82f89caab715907e3ff4d2c7b5e25c4 by documenting the ownership model on OpContext::trace_span.

The intended contract is that this field is a single-threaded parent handoff slot, not a concurrent current-span stack. The downstream search path copies the parent before entering parallel work: Cardinal snapshots the OpContext span before query tasks are created and then uses per-worker SearchContext::trace_span for nested spans inside those tasks, so workers do not push/pop op_context->trace_span concurrently.

foxspy (approved):
/lgtm

PR #98, opened 2026-06-24
enhance: add cache shard disk usage metric

sunby:
## Summary
- add optional shard attribution to cachinglayer translator metadata
- add `internal_cache_shard_disk_usage_bytes{data_type,shard}` gauge
- update `CacheSlot` loaded/unloaded transitions to increment/decrement loaded disk bytes by shard
- cover load and eviction accounting in cachinglayer cache slot test

Companion Milvus PR: https://github.com/milvus-io/milvus/pull/50733

### DIFF
--- a/include/cachinglayer/CacheSlot.h
+++ b/include/cachinglayer/CacheSlot.h
@@ -26,6 +26,7 @@
 #include <memory>
 #include <numeric>
 #include <optional>
+#include <string>
 #include <type_traits>
 #include <utility>
 #include <vector>
@@ -62,12 +63,16 @@ class CacheSlot final : public std::enable_shared_from_this<CacheSlot<CellT>> {
           cell_id_mapping_mode_(translator_->meta()->cell_id_mapping_mode),
           cell_data_type_(translator_->meta()->cell_data_type),
           storage_type_(translator_->meta()->storage_type),
+          shard_(translator_->meta()->shard),
           dlist_(dlist),
           evictable_(evictable),
           self_reserve_(self_reserve),
           storage_usage_tracking_enabled_(storage_usage_tracking_enabled),
           loading_timeout_(loading_timeout),
           warmup_loading_timeout_(warmup_loading_timeout) {
+        if (!shard_.empty()) {
+            shard_disk_usage_gauge_ = &monitor::cache_shard_disk_usage_bytes(cell_data_type_, shard_);
+        }
         cells_.reserve(translator_->num_cells());
         for (cid_t i = 0; i < static_cast<cid_t>(translator_->num_cells()); ++i) {
             cells_.push_back(std::make_unique<CacheCell>(this, i));
@@ -636,6 +641,9 @@ class CacheSlot final : public std::enable_shared_from_this<CacheSlot<CellT>> {
                         .Increment(loaded_size_.memory_bytes);
                     monitor::cache_loaded_bytes(slot_->cell_data_type_, StorageType::DISK)
                         .Increment(loaded_size_.file_bytes);
+                    if (slot_->shard_disk_usage_gauge_ != nullptr && loaded_size_.file_bytes > 0) {
+                        slot_->shard_disk_usage_gauge_->Increment(loaded_size_.file_bytes);
+                    }
                     monitor::cache_cell_loaded_count(slot_->cell_data_type_, slot_->storage_type_).Increment();
                 },
                 requesting_thread);
@@ -672,6 +680,9 @@ class CacheSlot final : public std::enable_shared_from_this<CacheSlot<CellT>> {
                     .Decrement(loaded_size_.memory_bytes);
                 monitor::cache_loaded_bytes(slot_->cell_data_type_, StorageType::DISK)
                     .Decrement(loaded_size_.file_bytes);
+                if (slot_->shard_disk_usage_gauge_ != nullptr && loaded_size_.file_bytes > 0) {
+                    slot_->shard_disk_usage_gauge_->Decrement(loaded_size_.file_bytes);
+                }
                 LOG_TRACE("[MCL] CacheSlot Cell unloaded: key={}, size={}", key(), loaded_size_.ToString());
                 loaded_size_ = {0, 0};  // reset loaded_size_ to 0,0 to avoid double refund from dlist_
             }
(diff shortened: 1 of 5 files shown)


sparknack (commented):
Moved these review notes to inline comments on the corresponding diff lines.

src/cachinglayer/Metrics.cpp
```diff
@@ -57,4 +60,30 @@ DEFINE_PROMETHEUS_HISTOGRAM_METRIC_WITH_DATA_TYPE_AND_LOCATION(internal_cache_ce
...
+            return "other";
+    }
+    return "unknown";
+}
+
+}  // namespace
+
+prometheus::Gauge&
+cache_shard_disk_usage_bytes(CellDataType type, const std::string& shard) {
+    return internal_cache_shard_disk_usage_bytes_family.Add({{"data_type", CellDataTypeLabel(type)}, {"shard", shard}});
```
sparknack: This dynamic label needs an explicit lifecycle. `Family::Add` keeps a time series until `Remove`, so a long-running QueryNode will retain one `{data_type, shard}` series for every insert channel it has ever loaded, even after the gauge returns to 0. Please add a ref-counted handle that removes the gauge when the last slot for that shard is gone, or avoid using a dynamic shard label here.

include/cachinglayer/Translator.h
```diff
@@ -43,15 +44,19 @@ struct Meta {
     // loads happen. The real resource usage is bounded by loading_pool_size * cell_size.
     // If not set, no capping is applied (existing behavior).
     std::optional<LoadingOverheadConfig> loading_overhead;
+    // Optional shard identifier for per-shard cache disk usage metrics.
+    // Empty means this translator is not attributed to any shard.
+    std::string shard;
```
sparknack: This puts Milvus query shard / insert-channel semantics into the generic cachinglayer `Meta`, and the field is public mutable state. I would prefer an explicit metric-attribution type or handle with documented value domain, lifecycle, and cardinality limit, or keeping this attribution in the Milvus layer instead of common caching metadata.

src/cachinglayer/Metrics.cpp
```diff
@@ -57,4 +60,30 @@ DEFINE_PROMETHEUS_HISTOGRAM_METRIC_WITH_DATA_TYPE_AND_LOCATION(internal_cache_ce
...
+        case CellDataType::VECTOR_INDEX:
+            return "vector_index";
+        case CellDataType::SCALAR_FIELD:
+            return "scalar_field";
+        case CellDataType::SCALAR_INDEX:
+            return "scalar_index";
+        case CellDataType::OTHER:
+            return "other";
+    }
+    return "unknown";
```
sparknack: Please keep the existing fail-fast behavior for unknown `CellDataType`. The other metric helpers throw on unknown enum values; returning `"unknown"` here would silently hide bad metadata or a future enum addition under a normal-looking label.

include/cachinglayer/CacheSlot.h
```diff
@@ -62,12 +63,16 @@ class CacheSlot final : public std::enable_shared_from_this<CacheSlot<CellT>> {
...
           cell_data_type_(translator_->meta()->cell_data_type),
           storage_type_(translator_->meta()->storage_type),
+          shard_(translator_->meta()->shard),
           dlist_(dlist),
           evictable_(evictable),
           self_reserve_(self_reserve),
           storage_usage_tracking_enabled_(storage_usage_tracking_enabled),
           loading_timeout_(loading_timeout),
           warmup_loading_timeout_(warmup_loading_timeout) {
+        if (!shard_.empty()) {
```
sparknack: Because empty shard skips this metric entirely, `internal_cache_shard_disk_usage_bytes` cannot be summed and reconciled with `internal_cache_loaded_bytes{location="disk"}`. Direct `Manager::ChargeLoadedResource` paths are also unattributed. Please either make the name/help text clear that this is only attributed cache-slot disk bytes, or define an explicit representation for unattributed usage.

test/test_cachinglayer/test_cache_slot.cpp
```diff
@@ -354,6 +414,31 @@ TEST_F(CacheSlotTest, PinMultipleCellsSuccess) {
...
+    {
+        auto op_ctx = std::make_unique<milvus::OpContext>();
+        auto accessor = cache_slot->PinCellsDirect(op_ctx.get(), {0, 1});
+        ASSERT_NE(accessor, nullptr);
+        ASSERT_NE(accessor->get_ith_cell(0), nullptr);
+        ASSERT_NE(accessor->get_ith_cell(1), nullptr);
+        EXPECT_EQ(monitor::cache_shard_disk_usage_bytes(CellDataType::SCALAR_FIELD, kShard).Value(), before + 384);
+    }
+
+    ASSERT_TRUE(cache_slot->ManualEvictAll());
```
sparknack, line 455: This covers the load + manual eviction happy path, but the risky accounting paths are still untested. Please also cover two slots sharing the same shard where one slot is released while the other remains loaded, and destroying a `CacheSlot` without manual eviction to verify the destructor / `CacheCell::mark_unload` path refunds the shard gauge.

src/cachinglayer/Metrics.cpp
```diff
@@ -57,4 +67,153 @@ DEFINE_PROMETHEUS_HISTOGRAM_METRIC_WITH_DATA_TYPE_AND_LOCATION(internal_cache_ce
...
+struct CacheShardDiskUsageMetricEntry {
+    ShardDiskUsageMetricKey key;
+    prometheus::Gauge* gauge{nullptr};
+};
+
+CacheShardDiskUsageMetricHandle::CacheShardDiskUsageMetricHandle(std::shared_ptr<CacheShardDiskUsageMetricEntry> entry)
+    : entry_(std::move(entry)) {
+}
+
+CacheShardDiskUsageMetricHandle::~CacheShardDiskUsageMetricHandle() = default;
```
yhmo, line 132: Medium — the last `CacheShardDiskUsageMetricHandle` now just drops its shared_ptr, so this `{data_type, shard}` series stays registered in `internal_cache_shard_disk_usage_bytes_family` until some later call happens to hit `collect_cache_shard_disk_usage_stats()` or `cache_shard_disk_usage_bytes_value()`. In production there is no normal cleanup caller here, which means a long-running node can retain one stale Prometheus series per shard it has ever loaded even after the gauge has been refunded to 0. The new tests also mask this because `ScalarFieldShardDiskUsage()` calls `collect_cache_shard_disk_usage_stats()` before asserting teardown. Please remove the family entry when the last handle dies instead of relying on a later helper call to GC it.

PR #102, opened 2026-07-01
enhance: add retriable StorageTransientError(2045) error code

czs007:
## What

Add a retriable storage transient fallback error code `StorageTransientError = 2045`, paired with the existing permanent `StorageError = 2044`, to `enum ErrorCode` in `include/common/EasyAssert.h`.

## Why

Producers (milvus-storage) and the milvus segcore boundary need a generic **retriable** storage code so a transient object-storage IO / throttling / timeout failure stays retriable, instead of collapsing into the non-retriable `StorageError`. The two codes are a pair — each carries exactly one retry verdict and they must never be conflated.

## Part of

The segcore error-classification effort: milvus-io/milvus#50903 (design). Consumed by the milvus-storage `ToSegcoreErrorCode` mapping and the milvus segcore code table.

### DIFF
--- a/include/common/EasyAssert.h
+++ b/include/common/EasyAssert.h
@@ -71,8 +71,16 @@ enum ErrorCode {
     TextIndexNotFound = 2041,
     InvalidParameter = 2042,
     InsufficientResource = 2043,
-    // milvus-storage related error code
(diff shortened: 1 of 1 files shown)


foxspy (approved):
/lgtm

PR #105, opened 2026-07-17
enhance: track loading overhead per dimension

sparknack:
- track memory and file loading overhead with independent group caps
- pass unconfigured dimensions through to DList admission
- preserve the legacy shared-group constructor by mapping it to both dimensions, so existing Cardinal callers remain source-compatible
- cover shared memory caps, temporary disk rejection, and legacy config registration

Refs: https://github.com/milvus-io/milvus/pull/51403

### DIFF
--- a/include/cachinglayer/LoadingOverheadTracker.h
+++ b/include/cachinglayer/LoadingOverheadTracker.h
@@ -30,7 +30,8 @@ namespace milvus::cachinglayer {
 //
 // The total loading overhead reserved from DList for each configured dimension is capped:
 //   DList dimension reservation = min(sum_of_overhead, dimension_UB)
-// An unconfigured dimension passes through unchanged.
+// An unconfigured dimension passes through unchanged. INT64_MAX is an explicit
+// unlimited upper bound; re-registering a group can only increase its upper bound.
 //
 // Each Reserve/Release call returns the incremental delta to apply to DList.
 // The tracker directly tracks `overhead_reserved` (actual amount of overhead currently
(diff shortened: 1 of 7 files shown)


include/cachinglayer/LoadingOverheadTracker.h
```diff
@@ -148,50 +127,139 @@ class LoadingOverheadTracker {
...
+    };
+
+    uint64_t
+    registerDimensionGroup(std::unordered_map<std::string, uint64_t>& name_to_handle,
+                           const LoadingOverheadDimensionConfig& config, const char* dimension) {
+        auto it = name_to_handle.find(config.group);
+        if (it != name_to_handle.end()) {
+            auto& state = dimension_group_state_[it->second];
+            state.ref_count++;
+            if (state.upper_bound == std::numeric_limits<int64_t>::max()) {
```
liliu-z: registerDimensionGroup replaces an existing dimension's upper_bound whenever it equals INT64_MAX (lines 160-161), treating a per-dimension INT64_MAX as "unconfigured". The legacy code only replaced the cap when the whole ResourceUsage equaled kUnlimited {INT64_MAX, INT64_MAX} and otherwise took the per-dimension max, so it never lowered an unlimited dimension. Through the shared-group compatibility overload this diverges: registering {INT64_MAX, 0} then {256 MiB, 0} on one group used to keep the memory cap unlimited but now silently lowers it to 256 MiB, so overhead above the finite cap is dropped from DList accounting and memory can overcommit. It is order-dependent ({finite} then {INT64_MAX} still yields unlimited), and matches the milvus load-transient group where an unlimited and a finite config share a key. Restore the whole-ResourceUsage sentinel in the compatibility path, or explicitly document and add a test for the behavioral break.

PR #108, opened 2026-07-19
fix: refresh cache loading overhead bounds

sparknack:
## Summary
- re-read translator loading-overhead config before every CacheSlot load
- track each live registration cap and use the maximum as the shared group bound
- support safe bound increases and decreases with exact failed-reservation rollback
- refresh tracker state without serializing on the DList-wide lock

## Contract
- configured dimensions and group names stay stable for the CacheSlot lifetime
- use INT64_MAX when a dimension starts uncapped but may become capped later

## Test Plan
- `cmake --build . -j8`
- `./build/test/test_cachinglayer/cachinglayer_test` (143 passed)
- `./build/test/all_tests` (112 passed)

Refs: https://github.com/milvus-io/milvus/pull/51403

### DIFF
--- a/include/cachinglayer/CacheSlot.h
+++ b/include/cachinglayer/CacheSlot.h
@@ -82,8 +82,8 @@ class CacheSlot final : public std::enable_shared_from_this<CacheSlot<CellT>> {
         monitor::cache_cell_count(cell_data_type_, storage_type_).Increment(translator_->num_cells());
         // Register after all potentially-throwing operations, so that if the constructor
         // fails, we don't leak a ref_count (destructor won't run for incomplete objects).
-        if (auto& lo = translator_->meta()->loading_overhead) {
-            overhead_handle_ = dlist_->RegisterLoadingOverhead(*lo);
+        if (auto loading_overhead = translator_->loading_overhead_config()) {
+            overhead_handle_ = dlist_->RegisterLoadingOverhead(*loading_overhead);
         }
     }
 
@@ -441,6 +441,9 @@ class CacheSlot final : public std::enable_shared_from_this<CacheSlot<CellT>> {
         //   pass through unchanged. The tracker returns the combined incremental DList delta.
         std::vector<cid_t> loading_cids;
         try {
+            if (auto loading_overhead = translator_->loading_overhead_config()) {
+                dlist_->RefreshLoadingOverheadUpperBound(overhead_handle_, *loading_overhead);
+            }
             auto start = std::chrono::steady_clock::now();
 
             loading_cids = std::vector<cid_t>(cids.begin(), cids.end());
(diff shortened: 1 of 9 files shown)


include/cachinglayer/CacheSlot.h
```diff
@@ -441,6 +441,9 @@ class CacheSlot final : public std::enable_shared_from_this<CacheSlot<CellT>> {
         //   pass through unchanged. The tracker returns the combined incremental DList delta.
         std::vector<cid_t> loading_cids;
         try {
+            if (auto loading_overhead = translator_->loading_overhead_config()) {
+                dlist_->RefreshLoadingOverheadUpperBound(overhead_handle_, *loading_overhead);
```
liliu-z, line 445: RunLoad refreshes only the loading slot's own overhead_handle_, but a group's effective upper bound is the max over every live registration's contribution (upper_bound = upper_bound_ref_counts.rbegin()->first). A slot that finished loading and stays resident never re-runs RunLoad, so its contribution is never refreshed; if the shared config decreases, that stale higher contribution pins the group bound. For the documented pattern of registering INT64_MAX at construction for a dimension that may later become capped (Translator.h), a single resident slot keeps the group max at INT64_MAX, so the group cap stays effectively unenforced (loading overhead is then bounded only by the global DList limit, not the configured group cap) until that slot reloads or unregisters. Refresh all live registrations on a config change, or track the group's target bound centrally, rather than only the caller's handle.

include/cachinglayer/LoadingOverheadTracker.h
```diff
@@ -120,6 +129,30 @@ class LoadingOverheadTracker {
         return getUpperBoundLocked(handle);
     }
 
+    // Refresh a registration without changing its ref count. Dimensions and
+    // group names are fixed at registration; only each registration's upper-bound
+    // contribution is updated. The group's effective bound remains the maximum
+    // contribution from all live registrations.
+    void
+    RefreshUpperBound(uint64_t handle, const LoadingOverheadConfig& config) {
```
liliu-z, line 137: RefreshUpperBound updates a group's upper bound but never re-evaluates the DList waiting queue. When a refresh lowers a group's cap, a blocked reservation for that same group needs a smaller overhead delta (reserveDimension caps delta at min(sum, upper_bound) - overhead_reserved) and may become admissible without any resource being freed, yet nothing schedules handleWaitingRequests. With background eviction disabled (the EvictionConfig default) and no other Release/touch activity, such a waiter can stall until its timeout even though it is now satisfiable. Schedule a handleWaitingRequests pass when a refresh yields an effective bound decrease; it must be scheduled asynchronously (e.g. on the event-base thread) to preserve this PR's goal of refreshing without taking the DList list lock.

yhmo:
@Mergifyio refresh

yhmo (commented):
@dcoapp recheck

PR #110, opened 2026-07-22
enhance: support dynamic loading overhead groups

sparknack:
## What changed
- replace tracker registrations and numeric handles with independently created shared LoadingOverheadGroup handles
- support Passthrough, Budget, and Executor policies with runtime policy replacement
- serialize Group binding, admission, rollback, release, and policy updates through DList
- bind Translator metadata to opaque Group handles and per-binding maximum runtime-unit bounds
- reconcile expansion on Reserve and drain tightening on Release without pending policy state

Supersedes #109.

Related Milvus integration: https://github.com/milvus-io/milvus/pull/51403

## Test
- cmake --build build --target all_tests cachinglayer_test -j2
- source build/conanrun.sh && ctest --test-dir build --output-on-failure (2/2 passed)
- clang-format --dry-run --Werror on changed C++ files
- git diff --check


src/cachinglayer/lrucache/DList.cpp
```diff
@@ -35,35 +36,89 @@ ClampNonNegative(std::atomic<ResourceUsage>& counter, LogFn&& log_fn) {
...
+    }
+}
+
+LoadingOverheadUpdateResult
+DList::UpdateLoadingOverheadGroup(const LoadingOverheadGroupHandle& group, LoadingOverheadPolicy policy) {
+    std::vector<std::unique_ptr<WaitingRequest>> to_destroy;
+    LoadingOverheadUpdateResult result;
+    {
+        std::unique_lock<std::mutex> lock(list_mtx_);
+        result = group ? group->updatePolicy(std::move(policy)) : LoadingOverheadUpdateResult::kInvalidArgument;
```
buqian-zilliz, line 116: **[P1] Charge expanded group targets before returning applied**

`UpdateLoadingOverheadGroup()` changes the policy and returns `kApplied` without charging the positive target delta for demand that has already passed DList admission. Under `Budget(100)`, two 100-byte loads can both pass DList while the second waits on the runtime budget because its group delta is zero; after this method applies `Budget(200)`, increasing the real budget can wake that load while `total_loading_size_` still accounts for only 100 bytes until some unrelated future reserve. This bypasses the cache and physical limits. Admit the target growth transactionally before returning `kApplied`, and reject or roll back the policy when the extra reservation cannot fit.

src/cachinglayer/LoadingOverhead.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+int64_t
+LoadingOverheadGroup::release(int64_t overhead) noexcept {
+    sum_of_overhead_ -= overhead;
+    if (sum_of_overhead_ < 0) {
+        LOG_ERROR("[MCL] LoadingOverheadGroup Release: sum_of_overhead < 0");
+        sum_of_overhead_ = 0;
+    }
+
+    const auto delta = std::max(overhead_reserved_ - computeTarget(), int64_t{0});
```
buqian-zilliz, line 131: **[P1] Keep pre-shrink work reserved until it drains**

`release()` immediately reconciles against the replacement policy, so a budget shrink can free reservation that still protects outstanding acquisitions. With four active 100-byte acquisitions, `sum_of_overhead_ = overhead_reserved_ = 400`; after lowering the real budget and policy to 200, the first 100-byte completion leaves 300 bytes in flight but this calculation releases 200 and leaves DList accounting for only 200. The budget update does not revoke existing acquisitions, so the remaining work can exceed the cache limit. Retain the old effective target or generation until pre-update work drains, or otherwise cap each drain to the reservation made releasable by the completed work.

src/cachinglayer/lrucache/DList.cpp
```diff
@@ -820,16 +943,17 @@ DList::handleWaitingRequests() {
...
-            auto [ok, unscaled] =
-                reserveResourceInternalWithTracker(request_ptr_ref->loaded, request_ptr_ref->overhead,
-                                                   request_ptr_ref->overhead_handle, loading_overhead_tracker_.get());
-            fulfilled = ok;
-            actual = unscaled;
+            auto attempt = reserveResourceInternalWithOverhead(request_ptr_ref->loaded, request_ptr_ref->overhead,
+                                                               request_ptr_ref->loadingOverheadConfig());
+            fulfilled = attempt.result.success;
+            actual = attempt.result.reserved;
+            attempted_requirement = attempt.required_size;
```
buqian-zilliz: **[P2] Reheap waiters when their requirement changes**

`attempt.required_size` is recomputed after group state or policy changes, but it is copied only to this local variable; the request remains ordered by its original `required_size`. For indefinite waiters, which share the same deadline, a request initially keyed at 5 bytes can grow to 20 bytes and remain ahead of an 8-byte request. With 10 bytes available, the top request fails and the loop breaks, so the 8-byte request can wait forever even though it fits. Reinsert the waiter when its requirement changes, or continue scanning peers with the same deadline before stopping.

include/cachinglayer/CacheSlot.h
```diff
@@ -78,13 +79,11 @@ class CacheSlot final : public std::enable_shared_from_this<CacheSlot<CellT>> {
         for (cid_t i = 0; i < static_cast<cid_t>(translator_->num_cells()); ++i) {
             cells_.push_back(std::make_unique<CacheCell>(this, i));
         }
+        if (loading_overhead_config_) {
+            dlist_->BindLoadingOverheadGroups(*loading_overhead_config_);
```
buqian-zilliz, line 83: **[P2] Roll back group bindings when construction fails**

`BindLoadingOverheadGroups()` commits shared membership before the constructor is complete. The memory binding can succeed and the file binding's `std::multiset` insertion can then throw, and after this call the metric helpers can also throw for an invalid enum. Because a throwing constructor never runs `~CacheSlot()`, `binding_count_` and `runtime_unit_bounds_` remain inflated, corrupting future Budget or Executor bounds or making updates permanently incompatible. Make the two-dimensional bind transactional and keep an RAII unbind guard until construction commits.

src/cachinglayer/lrucache/DList.cpp
```diff
@@ -224,41 +271,105 @@ DList::reserveResourceInternal(const ResourceUsage& size) {
...
+            rollbackLoadingOverhead(*config, overhead, delta);
         }
     };
 
-    auto [success, _] = reserveResourceInternalImpl(actual_size, rollback);
+    const auto unscaled = loaded + delta;
+    // FIXME: Fractional scaling is not additive. When Group Reserve and Release
+    // partition the same logical delta differently (for example after a policy
+    // update), per-transition rounding can make total_loading_size_ drift.
+    const auto scaled = unscaled * eviction_config_.loading_resource_factor;
```
buqian-zilliz: **[P3] Keep scaled group accounting additive**

`ResourceUsage::operator*` rounds each transition separately, while reserve and release can partition the same group delta differently. With a factor of 1.5 and a fixed bound of 2 bytes, requests with overheads 1 and 2 charge rounded deltas of 2 + 2 bytes; if the 1-byte request finishes first, the releases are 0 and `round(2 * 1.5) = 3`, leaving a permanent phantom byte in `total_loading_size_`. Repeated overlapping loads accumulate the drift and eventually cause false admission failures. Track the scaled aggregate target or the exact scaled amount charged instead of rescaling repartitioned deltas.

src/cachinglayer/lrucache/DList.cpp
```diff
@@ -35,35 +36,114 @@ ClampNonNegative(std::atomic<ResourceUsage>& counter, LogFn&& log_fn) {
...
+}
+
+}  // namespace
+
+LoadingOverheadGroupHandle
+DList::CreateLoadingOverheadGroup(LoadingOverheadDimension dimension, LoadingOverheadPolicy policy) {
+    if (dimension != LoadingOverheadDimension::kMemory && dimension != LoadingOverheadDimension::kFile) {
+        return nullptr;
+    }
+    return LoadingOverheadGroupHandle(new LoadingOverheadGroup(dimension, std::move(policy)));
```
buqian-zilliz, line 60: **[P2] Keep each group in one DList accounting domain**

`LoadingOverheadGroup` does not record which `DList` created it, while binding and update validation check only the resource dimension. The same handle can therefore be reused by multiple lists: with `Fixed(100)`, A's first 100-byte load charges A and B's load charges zero; if A releases first and B last, A retains a phantom 100-byte reservation while B subtracts one it never held. Concurrent use also mutates the group under different mutexes. Store and validate the owning `DList`, or give the group a single lock and accounting owner.

src/cachinglayer/lrucache/DList.cpp
buqian-zilliz: This issue still exists. `UpdateLoadingOverheadGroup()` still only swaps the policy and retries queued requests; it never charges the positive target delta for already-admitted work, so expanding the external Budget/TP can wake that work while `total_loading_size_` remains stale.

src/cachinglayer/LoadingOverhead.cpp
buqian-zilliz: This issue still exists. `release()` still reconciles immediately against the replacement policy, so the first completion after a tightening can release reservation that is still needed for pre-update in-flight acquisitions.

src/cachinglayer/lrucache/DList.cpp
buqian-zilliz: This issue still exists. The new reinsert path only refreshes the waiter currently at the heap top; a non-top waiter whose requirement shrinks can remain hidden behind an unchanged, unsatisfiable top waiter and wait indefinitely even though it now fits.

src/cachinglayer/lrucache/DList.cpp
sparknack: This is intentional. `UpdateLoadingOverheadGroup()` provides eventual, not transactional, reconciliation. The owner updates the Group before expanding the actual Budget/TP; already-admitted work may temporarily diverge from `total_loading_size_` until later Reserve/Release activity reconciles it. A strict DList hard-limit guarantee during this reconfiguration window is intentionally out of contract, avoiding rejected runtime updates and pending policy state. The ordering is documented in `Manager.h` and covered by `PolicyExpansionReconcilesOnNextReserve`, so I am leaving this unchanged.

src/cachinglayer/LoadingOverhead.cpp
sparknack: This is intentional. The owner tightens the actual Budget/TP first, then updates the Group. Existing acquisitions drain naturally, and DList accounting may temporarily diverge from physical in-flight usage during that transition. We explicitly chose not to retain an old policy generation or pending target. The ordering is documented in `Manager.h`, and the behavior is covered by `PolicyTighteningAppliesImmediately` and `PolicyTighteningDrainsExistingReservationWithInflightWork`, so I am leaving this unchanged.

include/cachinglayer/CacheSlot.h
sparknack: Fixed in bb87b63. `BindLoadingOverheadGroups()` validates both dimensions first and rolls back the memory bind if the file bind throws. `CacheSlot` also keeps an unbind guard until construction completes, covered by `ConstructorFailureDoesNotLeakGroupBinding`.

src/cachinglayer/lrucache/DList.cpp
sparknack: Fixed in 5a08bc0. Finite waiters now ceiling-align their deadlines to 10 ms buckets, and `handleWaitingRequests()` scans every waiter in the blocked bucket once. `PolicyShrinkRechecksNonTopWaiterInSameDeadlineBucket` covers a shrinking non-top waiter hidden behind an unsatisfiable top waiter.

src/cachinglayer/lrucache/DList.cpp
sparknack: Fixed in 78b720a. Request-local usage is scaled separately, while Group accounting applies `scale(aggregate_after) - scale(aggregate_before)` on reserve and the inverse on release. This keeps rounding additive across different partition and release orders; the 1.5 factor case is covered by `FractionalLoadingFactorKeepsGroupAggregateAdditive`.

src/cachinglayer/lrucache/DList.cpp
sparknack: This is outside the supported accounting contract. Production has a single `Manager`-owned `DList`; Group handles are created and used within that DList. Cross-DList sharing is only reachable through direct internal or test misuse, so I am leaving the owner field out.

src/cachinglayer/lrucache/DList.cpp
```diff
@@ -35,35 +36,125 @@ ClampNonNegative(std::atomic<ResourceUsage>& counter, LogFn&& log_fn) {
...
+    auto attempt = reserveResourceInternalWithOverhead(loaded, overhead, config);
+    if (attempt.result.success) {
+        return folly::makeSemiFuture(attempt.result);
     }
-    auto [success, actual] =
-        reserveResourceInternalWithTracker(loaded, overhead, overhead_handle, loading_overhead_tracker_.get());
-    if (success) {
-        return folly::makeSemiFuture(actual);
+
+    if (!max_resource_limit_.load().CanHold(attempt.required_size)) {
```
buqian-zilliz: **[P2] Honor timeouts across dynamic policy changes**

`attempt.required_size` includes the Group delta under the current policy, but this branch completes the future before honoring a nonzero timeout. With a 100-byte limit and factor 1, a 60-byte local load plus 60 bytes of grouped overhead under `Fixed(60)` is rejected at 120 bytes even though tightening the Group to `Fixed(30)` before the deadline would reduce the requirement to 90 bytes; `UpdateLoadingOverheadGroup()` already retries waiters after such changes. Use only the policy-independent request-local floor for permanent rejection, and apply the same rule to the queued-request check at line 1048.

src/cachinglayer/lrucache/DList.cpp
```diff
@@ -35,35 +36,125 @@ ClampNonNegative(std::atomic<ResourceUsage>& counter, LogFn&& log_fn) {
...
+    if (config.memory.has_value()) {
+        if (config.memory->group) {
+            config.memory->group->unbind(LoadingOverheadDimension::kMemory, config.memory->max_runtime_unit);
+        } else {
+            LOG_ERROR("[MCL] LoadingOverheadGroup cannot unbind invalid memory Group");
+        }
+    }
+    if (config.file.has_value()) {
+        if (config.file->group) {
+            config.file->group->unbind(LoadingOverheadDimension::kFile, config.file->max_runtime_unit);
```
buqian-zilliz: **[P2] Wake waiters after lowering a binding bound**

`UnbindLoadingOverheadGroups()` can remove the largest `max_runtime_unit`, lowering a Budget or Executor Group target, but it returns without retrying the queue. If that binding belongs to an idle CacheSlot, member destruction produces no `ReleaseLoadingResource()` or `RefundLoadedResource()` notification; with background eviction disabled, an indefinite waiter whose recomputed delta now fits can remain blocked forever. Run `handleWaitingRequests()` after both unbinds and destroy completed requests after releasing `list_mtx_`.

src/cachinglayer/lrucache/DList.cpp
```diff
@@ -844,45 +1025,69 @@ DList::handleWaitingRequests() {
...
                 requests_to_destroy.push_back(std::move(request));
                 waiting_queue_.pop();
                 continue;
             }
-            LOG_DEBUG("[MCL] Request {} of size {} cannot be satisfied, breaking.", request_ptr_ref->request_id,
-                      request_ptr_ref->required_size.ToString());
-            // Cannot satisfy right now but may succeed later.
-            // The queue is ordered by deadline, so stop here.
-            break;
+            auto request = std::move(request_ptr_ref);
```
buqian-zilliz, line 1093: **[P2] Bound same-deadline waiter rescans**

Once one request cannot fit, this path removes, fully retries, and reinserts every peer with the same deadline while holding `list_mtx_`. All negative-timeout requests share `time_point::max()` and finite bursts share 10 ms buckets, so each release, cancellation, or policy update can perform O(N log N) heap work plus N physical-memory and LRU admission scans; draining N requests becomes O(N^2 log N) exactly under cache pressure. Reheap only waiters whose Group-derived key changed, or bound or partition the peer scan.

src/cachinglayer/lrucache/DList.cpp
sparknack: Fixed in a028bee. `UnbindLoadingOverheadGroups()` now retries the waiting queue after both dimensions are unbound and keeps completed `WaitingRequest` destruction outside `list_mtx_`. Covered by `UnbindingLargestRuntimeUnitWakesWaiters`.

src/cachinglayer/lrucache/DList.cpp
sparknack: Partially addressed in c4c75ed by removing 10 ms deadline bucketing. Finite waiters retain their exact deadlines and are scanned together only when those deadlines are exactly equal. Negative-timeout waiters still share `time_point::max()` and scan equal-deadline peers; that is an intentional correctness trade-off to avoid reintroducing non-top starvation without adding mutable waiter-index or queue state, so I am leaving the indefinite worst case unchanged in this PR.

src/cachinglayer/lrucache/DList.cpp
sparknack: Fixed in a028bee. Permanent-impossibility checks now use only `RequestLocalLoading(...) * loading_resource_factor` in both initial admission and queued retries, so policy-derived Group overhead can shrink before timeout. Covered by `PolicyDependentRequirementAboveCapacityWaitsForShrink` and `QueuedRequirementAboveCapacityWaitsForLaterShrink`.

src/cachinglayer/LoadingOverhead.cpp
```diff
@@ -0,0 +1,177 @@
...
+    if (requiresRuntimeUnitBound(policy) && hasMissingRuntimeUnit()) {
+        return LoadingOverheadUpdateResult::kIncompatiblePolicy;
+    }
+    policy_ = policy;
+    return LoadingOverheadUpdateResult::kApplied;
+}
+
+int64_t
+LoadingOverheadGroup::reserve(int64_t overhead) {
+    sum_of_overhead_ += overhead;
```
buqian-zilliz, line 103: **[P2] Check aggregate demand before adding it**

`reserve()` adds each non-negative estimate to the signed `int64_t` aggregate before applying the policy bound. Because a capped Group can admit a request whose estimate is near `INT64_MAX` while booking only the cap, a second positive request can overflow `sum_of_overhead_`; signed overflow is undefined behavior and can make `computeTarget()` collapse toward zero, silently under-accounting later reservations and releases. Use checked or wider arithmetic and reject an unrepresentable reservation before mutating the Group.

docs/design-docs/20260723-loading-overhead-management-and-accounting.md
```diff
@@ -0,0 +1,860 @@
...
+A policy update or another request's Group activity can change a waiter's required size while it is queued. The queue therefore cannot treat the original size key as permanent.
+
+During processing:
+
+- the current requirement is recomputed;
+- a permanently impossible request whose requirement exceeds DList capacity fails immediately;
+- a failed request is reinserted with its refreshed requirement;
+- peers in the same deadline bucket are also examined so a stale or newly larger top request cannot hide a smaller request that now fits.
+
+Finite deadlines are rounded up into 10 ms buckets. Rounding never shortens the caller's timeout and lets DList fairly scan requests with effectively equivalent deadlines. Indefinite waiters share the maximum deadline and are scanned as one bucket.
```
buqian-zilliz: **[P3] Update the documented waiter deadline model**

This paragraph says finite deadlines are rounded into 10 ms buckets, but the current code stores an exact `steady_clock::now() + timeout` value and only continues scanning while the next deadline is exactly equal. Readers will therefore expect finite peers to be reconsidered together when they are not, and the stated performance and fairness model no longer matches the implementation. Document the exact-deadline behavior, including the shared `time_point::max()` deadline for indefinite waiters, or restore the bucketing logic.

src/cachinglayer/lrucache/DList.cpp
```diff
@@ -820,16 +1005,17 @@ DList::handleWaitingRequests() {
...
         bool fulfilled = false;
         ResourceUsage actual{};
+        auto attempted_requirement = request_ptr_ref->required_size;
         if (request_ptr_ref->use_resource_promise) {
-            auto [ok, unscaled] =
-                reserveResourceInternalWithTracker(request_ptr_ref->loaded, request_ptr_ref->overhead,
-                                                   request_ptr_ref->overhead_handle, loading_overhead_tracker_.get());
-            fulfilled = ok;
-            actual = unscaled;
+            auto attempt = reserveResourceInternalWithOverhead(request_ptr_ref->loaded, request_ptr_ref->overhead,
```
buqian-zilliz: **[P2] Catch overflow from queued Group retries**

A Group-aware waiter can pass the aggregate check when enqueued, then other capped or zero-delta admissions can raise the Group sum so this retry throws `std::overflow_error`. `handleWaitingRequests()` has no exception boundary: callers may already have committed policy, binding, or release changes, background-thread or destructor callers can terminate the process, and any peers already moved to `deferred_requests` are destroyed while `waiting_requests_map_` still holds their raw pointers. Catch this per waiter, erase and complete it cleanly or restore it to the heap, then continue so the outer transition and queue ownership remain consistent.

src/cachinglayer/LoadingOverhead.cpp
```diff
@@ -0,0 +1,188 @@
...
+    }
+}
+
+void
+LoadingOverheadGroup::bind(const std::optional<int64_t>& max_runtime_unit) {
+    if (max_runtime_unit.has_value()) {
+        runtime_unit_bounds_.insert(max_runtime_unit.value());
+        max_runtime_unit_ = std::max(max_runtime_unit_, max_runtime_unit.value());
+    }
+    ++binding_count_;
```
zhengbuqian, line 53: what does it mean when max_runtime_unit is empty? why do we still `++binding_count_;`?

src/cachinglayer/LoadingOverhead.cpp
```diff
@@ -0,0 +1,188 @@
...
+    LOG_DEBUG("[MCL] LoadingOverheadGroup bound {} Group (bindings={})", DimensionName(dimension_), binding_count_);
+}
+
+void
+LoadingOverheadGroup::unbind(LoadingOverheadDimension dimension,
```
zhengbuqian: why does only `unbind` need `LoadingOverheadDimension dimension` but not `bind`?

include/cachinglayer/LoadingOverhead.h
```diff
@@ -12,37 +12,207 @@
...
 
-// A missing dimension is not group-capped. Its loading overhead passes through
-// unchanged and is still checked against the DList resource limit.
-struct LoadingOverheadConfig {
-    LoadingOverheadConfig() = default;
+/** @brief Immutable Group policy configuration. */
+class LoadingOverheadPolicy {
+ private:
+    enum class Kind {
+        kFixed,
```
zhengbuqian: kFixed is no longer sued, consider removing it

include/cachinglayer/CacheSlot.h
```diff
@@ -78,13 +79,17 @@ class CacheSlot final : public std::enable_shared_from_this<CacheSlot<CellT>> {
...
+        }
+        auto unbind_on_failure = folly::makeGuard([this]() {
```
zhengbuqian, line 85: seems to me this guards only metrics update failure? is it necessary?

src/cachinglayer/LoadingOverhead.cpp
sparknack: An empty `max_runtime_unit` is allowed for a Passthrough binding. `binding_count_` tracks every attached binding, while `runtime_unit_bounds_` tracks only bindings that supplied a bound. Their count difference blocks a later switch to Budget/Executor. `runtime_unit_bounds_` is a multiset, so duplicate values count separately.

src/cachinglayer/LoadingOverhead.cpp
sparknack: Agreed. Group dimension is immutable and `BindLoadingOverheadGroups()` already validates it before mutation, so the `unbind` parameter was redundant. I removed it in follow-up `8eb5639`; the next push will include it.

include/cachinglayer/LoadingOverhead.h
sparknack: Agreed. I removed `kFixed`/`Fixed()`, updated the tests, and removed the English design-doc references in `79f9b43`. The next push will include it.

include/cachinglayer/CacheSlot.h
sparknack: It protects the post-bind constructor tail, currently the two metric updates. Those can throw for an invalid `CellDataType`; `ConstructorFailureDoesNotLeakGroupBinding` verifies the guard unbinds the Group so a later bounded-policy update is allowed. I will keep it.

src/cachinglayer/lrucache/DList.cpp
```diff
@@ -820,16 +1005,29 @@ DList::handleWaitingRequests() {
...
         ResourceUsage actual{};
+        auto attempted_requirement = request_ptr_ref->required_size;
         if (request_ptr_ref->use_resource_promise) {
-            auto [ok, unscaled] =
-                reserveResourceInternalWithTracker(request_ptr_ref->loaded, request_ptr_ref->overhead,
-                                                   request_ptr_ref->overhead_handle, loading_overhead_tracker_.get());
-            fulfilled = ok;
-            actual = unscaled;
+            try {
+                auto attempt = reserveResourceInternalWithOverhead(request_ptr_ref->loaded, request_ptr_ref->overhead,
```
buqian-zilliz, line 1021: **[P2] Reheap changed waiter requirements before admission**

`handleWaitingRequests()` retries and commits the heap top before replacing its stale `required_size`; the key is updated and reinserted only after a failed attempt. With two indefinite waiters, A can be queued at 20 bytes ahead of B at 50 while A's Group is saturated; when the last active load in that Group releases 50 bytes of loaded data plus 40 bytes of Group booking, A's current requirement becomes 60 and usage drops to 10. This path admits A first and then cannot fit B, whereas refreshing the heap would admit the smaller 50-byte waiter. Add a non-mutating requirement refresh and reheap equal-deadline waiters before committing any reserve or eviction.

include/cachinglayer/LoadingOverhead.h
```diff
@@ -12,37 +12,198 @@
...
+     */
+    LoadingOverheadGroupHandle group;
+
+    /**
+     * @brief Optional conservative bound for one runtime unit from this binding.
+     *
+     * The Group caches the maximum value across attached bindings. A
+     * policy that bounds Budget acquisitions or executor tasks requires
+     * this value on every binding; binding or policy replacement is
+     * rejected otherwise. Fixed and Passthrough Groups do not require it. It
```
buqian-zilliz: **[P3] Remove the deleted Fixed policy from the public contract**

`LoadingOverheadPolicy::Fixed()` was removed, but this public binding contract still describes Fixed Groups, and the design document still lists fixed-policy support as a goal. Consumers are therefore told that a policy exists and has binding semantics even though the API cannot construct it. Remove the remaining Fixed references, including the design goal, or restore the policy.

src/cachinglayer/lrucache/DList.cpp
```diff
@@ -224,41 +304,127 @@ DList::reserveResourceInternal(const ResourceUsage& size) {
...
+    validateLoadingOverheadBinding(config.memory, LoadingOverheadDimension::kMemory);
+    validateLoadingOverheadBinding(config.file, LoadingOverheadDimension::kFile);
+
+    if (config.memory.has_value()) {
+        config.memory->group->validateReserve(overhead.memory_bytes);
+    }
+    if (config.file.has_value()) {
+        config.file->group->validateReserve(overhead.file_bytes);
+    }
+
```
zhengbuqian, line 335: those are redundant? we will check in `reserve`. if we can eliminate this, we can remove the `validateReserve` method.

zhengbuqian (approved):

src/cachinglayer/lrucache/DList.cpp
sparknack: The prechecks are intentional. `reserveLoadingOverhead()` mutates the memory Group before the file Group. If the memory reserve succeeds and the file reserve then rejects an overflow, the function has not returned a delta for the outer rollback path, leaving a partial mutation. Validating both dimensions before either `reserve()` preserves two-dimensional atomicity. The checks inside `reserve()` are defensive after this prevalidation, but the outer `validateReserve()` phase cannot be removed.

src/cachinglayer/lrucache/DList.cpp
sparknack: We are keeping this as best-effort ordering. `required_size` is a cached heap key; policy or aggregate changes can make it stale while queued. If the stale top still fits, DList admits it without rebuilding all equal-deadline peers. Strict current-size ordering would require repeated O(N) refresh and reheap work because each admission changes shared Group state. Capacity and admission safety are unchanged; only same-deadline size ordering is not strict. I have documented this contract in `handleWaitingRequests()` and the design document in the local follow-up.

PR #112, opened 2026-07-24
enhance: carry typed error codes on cachinglayer failure paths and observe untyped cgo exceptions

czs007:
Part of the error-handling hardening tracked in milvus-io/milvus#50903 (see also milvus-io/milvus#50768, which consumes both changes).

## 1. ListNode: loading-in-destructor is an invariant violation, not a cancel

`ListNode::~ListNode` finding a cell still in `LOADING` state is an internal invariant violation — the slot must not be torn down while a load is in flight. It previously threw a bare `std::runtime_error`, which collapses to `UnexpectedError(2001)` at the cgo boundary; now it throws a typed `SegcoreError(UnexpectedError)` explicitly, keeping the intent readable and the throw site consistent with the cancel paths that #111 already typed as `FollyCancel`.

## 2. Observer hook for untyped exceptions at the cgo boundary

`FailureCStatus` recovers the typed code via `dynamic_cast<SegcoreError>`; anything else (a raw `std::exception` from a third-party lib, a sliced rethrow) silently collapses to 2001 and is invisible in production.

This adds `RegisterUntypedCgoExceptionObserver(cb)`: `FailureCStatus` invokes the callback whenever it hits a non-SegcoreError exception. milvus registers a Prometheus counter (`internal_cgo_untyped_exception_total`) plus a rate-limited warn at `SegcoreInit`, so "untyped exception rate should trend to zero" becomes a measurable claim instead of a hope. No behavior change when no observer is registered; a single atomic load on the failure path only.

## Verification

- Full unit suite green on this branch (rebased on current master, incl. #111).
- Consumer side: milvus-io/milvus#50768 unit test `StorageErrorCode.UntypedCgoExceptionObserverFires` exercises the hook end-to-end through `FailureCStatus`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/common/EasyAssert.cpp
+++ b/src/common/EasyAssert.cpp
@@ -16,11 +16,34 @@
 
 #include "common/EasyAssert.h"
 
+#include <atomic>
 #include <boost/stacktrace.hpp>
 #include <iostream>
 #include <sstream>
 
 #include "fmt/format.h"
+
+namespace milvus {
+
+namespace {
+std::atomic<UntypedCgoExceptionObserver> untyped_cgo_exception_observer{nullptr};
+}  // namespace
+
+void
+RegisterUntypedCgoExceptionObserver(UntypedCgoExceptionObserver observer) {
+    untyped_cgo_exception_observer.store(observer, std::memory_order_release);
+}
+
+namespace impl {
+void
+NotifyUntypedCgoException(const char* what) {
+    if (auto observer = untyped_cgo_exception_observer.load(std::memory_order_acquire)) {
+        observer(what);
+    }
+}
+}  // namespace impl
+
+}  // namespace milvus
 
 namespace milvus::impl {
 
--- a/src/common/EasyAssert.cpp
+++ b/src/common/EasyAssert.cpp
@@ -36,9 +36,17 @@ RegisterUntypedCgoExceptionObserver(UntypedCgoExceptionObserver observer) {
 
 namespace impl {
 void
-NotifyUntypedCgoException(const char* what) {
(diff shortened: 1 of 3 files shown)


src/common/EasyAssert.cpp
```diff
@@ -16,12 +16,35 @@
...
+void
+RegisterUntypedCgoExceptionObserver(UntypedCgoExceptionObserver observer) {
+    untyped_cgo_exception_observer.store(observer, std::memory_order_release);
+}
+
+namespace impl {
+void
+NotifyUntypedCgoException(const char* what) {
+    if (auto observer = untyped_cgo_exception_observer.load(std::memory_order_acquire)) {
+        observer(what);
```
buqian-zilliz: **[P1] Prevent observer exceptions from escaping the cgo boundary**

`FailureCStatus` must complete the exception-to-`CStatus` conversion, but this synchronous metric/log observer is called without an exception barrier. If the observer throws, for example during allocation or formatting, the original failure is replaced and the exception can escape the cgo boundary, terminating the process instead of returning `UnexpectedError`. Make `NotifyUntypedCgoException` best-effort and `noexcept`, and catch all exceptions from the observer.

czs007: Good catch — fixed in 03e3e2e. `NotifyUntypedCgoException` is now `noexcept` and wraps the observer call in `catch (...)`: the observer is metrics/logging only, so a throwing observer must never replace the original failure or let an exception escape mid-conversion at the cgo boundary. `make test` green.

zhengbuqian (approved):

PR #113, opened 2026-07-28
enhance: Add service name differentiation for tracer

presburger:
This change improves trace identification when multiple Milvus clusters share the same Jaeger backend. The cluster identifier is passed through TraceConfig and prepended to the Segcore OpenTelemetry service name. As a result, the service name changes from segcore to <cluster-id>-segcore, making it easy to filter and distinguish traces from different clusters in Jaeger while preserving the existing tracing and exporter behavior.
Note: Once this commit is merged, we will submit a corresponding follow-up commit to Milvus to pass the cluster identifier into milvus-common when initializing and resetting the Segcore tracing configuration.

### DIFF
--- a/include/common/Tracer.h
+++ b/include/common/Tracer.h
@@ -31,6 +31,7 @@ struct TraceConfig {
     std::string otlpHeaders;
     bool oltpSecure;
 
+    std::string clusterID;
     int nodeID;
 };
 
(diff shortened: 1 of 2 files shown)


presburger:
@czs007 help review.

yhmo:
@Mergifyio refresh

yhmo (commented):
@dcoapp recheck

